PR #12, opened 2015-01-28
CircumscribedCircle, LinearBezier Curves, multi repeats arrows,CircleColorOrdering

fluddokt:


itdelatrisu:
Wow, this looks great!  Thank you so much for handling this!
I can do some style cleaning after merging, but do you mind writing (very!) short comments for all the functions and function parameters?  It'd make my life a lot easier.  (Also, I'll separate the subclasses into separate files later, just FYI.)

itdelatrisu:
Also, can you write some fallbacks instead of throwing errors?  One badly-drawn slider is way better than the program crashing.  For example, on this map, I get "trying to continue Beziers with less than 2 points": https://osu.ppy.sh/s/47846

PR #14, opened 2015-02-07
Fixed pom.xml

BOHverkill:
I have fixed the apostrophe in the arguments, so that I was able to built opsu.
I have tried that on Linux so I do not know if this rans also on Windowz.


itdelatrisu:
Ahh, thank you.  Sorry for being so careless. x-x

PR #15, opened 2015-02-09
MP3 Audio Streaming through modifying some of slick's openal files

fluddokt:
May rarely crash/not play on next music.
May have introduced other unknown bugs. 
This may be due to loading it async.
Also loading of ogg is not loaded async anymore as it seems it doesn't
work very well.

### DIFF
--- a/src/itdelatrisu/opsu/audio/MusicController.java
+++ b/src/itdelatrisu/opsu/audio/MusicController.java
@@ -82,8 +82,6 @@ public class MusicController {
 				trackLoader = new Thread() {
 					@Override
 					public void run() {
-						//Loading ogg async seems to screw up
-						//So does mp3, but much less
 						loadTrack(osu.audioFilename, osu.previewTime, loop);
 					}
 				};
@@ -94,9 +92,6 @@ public class MusicController {
 					@Override
 					public void run() {
 						loadTrack(osu.audioFilename, osu.previewTime, loop);
-						//convertMp3(osu.audioFilename);
-						//if (!Thread.currentThread().isInterrupted())
-						//	loadTrack(wavFile, osu.previewTime, loop);
 					}
 				};
 				trackLoader.start();
@@ -339,15 +334,18 @@ public class MusicController {
 
 		// TODO: properly interrupt instead of using deprecated Thread.stop();
 		// interrupt the conversion/track loading
-		if (isTrackLoading())
+		// Not sure if the interrupt does anything
+		// And the join kind of defeats the purpose of threading it.
+		// But is needed since bad things happen when OpenALStreamPlayer source is released asynchronously I think.
+		if (isTrackLoading()){
+			//trackLoader.stop();
+			trackLoader.interrupt();
 			try {
 				trackLoader.join();
 			} catch (InterruptedException e) {
-				// TODO Auto-generated catch block
 				e.printStackTrace();
 			}
-//			trackLoader.interrupt();
-			//trackLoader.stop();
+		}
 		trackLoader = null;
 
 		// delete temporary WAV file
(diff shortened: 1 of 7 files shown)


itdelatrisu:
This looks fantastic -- thanks again!!  I haven't done much testing yet, but everything feels really smooth so far.

I've also experienced the bugs/crashes fairly often, though.  For long songs like https://osu.ppy.sh/s/29157, sometimes clicking on them in the song menu will just produce noise, but it works when actually playing the map.  Is there maybe something wrong with the track position skipping?

I also got this crash once when going from the main menu to the song menu (so, when the first song loaded):
```
** Uncaught Exception! **
java.lang.ArrayIndexOutOfBoundsException: 580
	at javazoom.jl.decoder.LayerIIIDecoder.huffman_decode(LayerIIIDecoder.java:795)
	at javazoom.jl.decoder.LayerIIIDecoder.decode(LayerIIIDecoder.java:278)
	at javazoom.jl.decoder.LayerIIIDecoder.decodeFrame(LayerIIIDecoder.java:219)
	at javazoom.jl.decoder.Decoder.decodeFrame(Decoder.java:147)
	at org.newdawn.slick.openal.Mp3InputStream.read(Mp3InputStream.java:74)
	at org.newdawn.slick.openal.Mp3InputStream.read(Mp3InputStream.java:110)
	at org.newdawn.slick.openal.Mp3InputStream.read(Mp3InputStream.java:134)
	at org.newdawn.slick.openal.OpenALStreamPlayer.stream(OpenALStreamPlayer.java:265)
	at org.newdawn.slick.openal.OpenALStreamPlayer.update(OpenALStreamPlayer.java:238)
	at org.newdawn.slick.openal.SoundStore.poll(SoundStore.java:938)
	at org.newdawn.slick.Music.poll(Music.java:46)
	at org.newdawn.slick.GameContainer.updateAndRender(GameContainer.java:643)
	at itdelatrisu.opsu.Container.updateAndRender(Container.java:108)
	at org.newdawn.slick.AppGameContainer.gameLoop(AppGameContainer.java:411)
	at itdelatrisu.opsu.Container.start(Container.java:65)
	at itdelatrisu.opsu.Opsu.main(Opsu.java:160)
```

I'd love to merge this, but it'd be nice if it could be a little more stable first.  I can run more tests tomorrow, and let me know if there's anything else you want me to do.  Again, thanks so much!

itdelatrisu:
Great!  I'm not getting any errors anymore, but yeah, the memory issues are still pretty bad.  Is there anything else you still want to try before I merge this?

Also, I can do the cleanup like last time, but it'd be a big help if you could delete some of the commented code and etc. that doesn't work, and only leave the comments that you want to keep (i.e. anything potentially helpful) -- since I can't tell the difference.

itdelatrisu:
Okay, thanks so much!  There are still memory leaks, though they're not quite as bad -- if you ever figure out what's causing them, let me know!

Also, I'd be happy to help out with your Android port (it's really impressive!), so ask me if you need anything.  (I recently uploaded a higher-resolution icon for you to use: 484bb1b.)

fluddokt:
How do you find out it has memory leaks? I'm not really sure how to find this out.

And the Android port, I don't know, I'm not sure where I'm going with it.
Do you want to release it with your releases?
Do you want the launchers, they are in the directory above, so I'm not sure how to add it to git.
It also can run on desktop and theoretically on iOS since it is using libgdx.

itdelatrisu:
I haven't run any memory reports, but just looking at my Task Manager (in Windows 8), the memory usage goes up for every track played (before, it would reset for each new track).  I think the actual problem is that the music sources are not being cleared/closed somewhere *fast enough*, since the memory is freed for me after about 5-10 seconds (and calling `System.gc()`).  I'm noticing this when trying to delete songs in my `OsuGroupList.get().deleteSongGroup()` method (which calls `Utils.deleteToTrash()`) -- I can't delete songs immediately after they've stopped playing anymore.

I think the Android app is *way* better than osu!droid, and I'm sure a lot of people would use it if you (or we) posted about it on the osu! forums.  It still crashes a lot (especially with the downloads), so that would have to be taken care of first.  I don't know anything about phone development, but if there are any features you want me to add that would help for phones, I can certainly try (if I have time).  And sure, I'd love to release it with my releases once it's more stable!

PR #16, opened 2015-02-09
Implement XDG Base Directory Support for Linux

### LINKED ISSUE #13: Save beatmaps/config etc. XDG-conform
Currently all is stored in the working directory.

"**$XDG_DATA_HOME** defines the base directory relative to which user specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to **$HOME/.local/share** should be used. "

"**$XDG_CONFIG_HOME** defines the base directory relative to which user specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to **$HOME/.config** should be used."

Pls read this carefully:
http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

SuperNascher:
This Pull-Request adds the XDG Base Directory Specification for Linux.

This resolves #13

### DIFF
--- a/src/itdelatrisu/opsu/Options.java
+++ b/src/itdelatrisu/opsu/Options.java
@@ -37,7 +37,7 @@ import java.util.concurrent.TimeUnit;
  */
 public class Options {
 	
-	public static String SEPARATOR = System.getProperty("file.separator");
+	public static String SEPARATOR = File.separator;
 	private static String OS = System.getProperty("os.name").toLowerCase();
 	public static String ConfigDir = getDefaultConfigDir();
 	public static String DataDir = getDefaultDataDir();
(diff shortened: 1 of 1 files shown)


itdelatrisu:
I'll get to this soon, sorry for the delay.  I'm thinking that I'll have a switch in the config file to enable/disable this, and have this *off* by default (to be consistent with past behavior).

Is there any reason you used `System.getProperty("file.separator")` instead of just `/` (which works in all operating systems I've seen) or `File.separator` (the Java standard)?

SuperNascher:
There was no spezific reason for using ```
System.getProperty("file.separator")```
, but I prefer ```
File.separator```
 over the ```
/```.

itdelatrisu:
Added in ed2d305, with some modifications since I wanted this off by default.  Due to how JarSplice works, you'll have to uncomment a line in pom.xml when packaging opsu! (in Maven) to use these directories.  Thanks again!

SuperNascher:
Thank you, for adding this Pull-Request :+1:

PR #17, opened 2015-02-14
Reworked Spinner

PizzaLovers007:
Made the spinner feel more like the osu! spinner.

+Calculated the RPM as the average of the last 50 read RPM's


itdelatrisu:
Looks great.  Thanks so much for your help!

One really insignificant thing: `rpm` isn't being set for the Auto/Spun-Out mods.  I'll just add it to the TODO. :P

itdelatrisu:
FPS seems to be having an effect on spinner difficulty -- it's a lot harder to spin at 60fps vs. 240fps.  Could you look at it?

PizzaLovers007:
Alright I'll try to look into it this weekend

chong601:
It actually also affected when running opsu! at arbitrary FPS (for example 1000+FPS and yes that is possible). The spinner cannot be cleared at all even on the easiest ones. 

Also, I found that even i can spin 400+ on osu!, it doesn't seem to happen on opsu! even if i spin like a madman (regardless whether the fps is 60, 120 or 240).

itdelatrisu:
fluddokt fixed this in #37 -- thanks though!

PR #18, opened 2015-02-14
close mp3 file handles.

fluddokt:
Close file handles so files can be deleted.

Not sure if this fixed the memory leak.


itdelatrisu:
Fantastic, thanks as always!

PR #20, opened 2015-02-14
Visual Changes 

fluddokt:
Visual Changes …

mostly personal preferences though

Larger hit circles
Circle overlay after drawing circle
Fade in approach circles
Fade in sliders (I don't think it looks that bad)
Fade out hit lighting
Hit Error Bar
 Probably need an option for this.
Miss locations(I don't know)

Reverse Slider ball rotation on repeats

Fill bg image to screen while keeping aspec ratio
Added a few more resolutions


itdelatrisu:
This looks great!  I didn't even notice a lot of these things before, but everything looks so much better with your changes. o_o

The error bar actually seems pretty useful, but I'll have it disabled by default and throw a switch in the options menu afterwards.  Sometimes the ticks are really far off the screen, though -- [this](http://i.imgur.com/nb3vyTI.jpg), for example -- and I think it'd be better if you could keep them inside the bar.  Maybe also hide the bar during breaks?

What's the purpose of the miss locations?  I can't think of why that'd be helpful, so I'd rather not keep them (unless you have a reason to!).

itdelatrisu:
Great, thanks so much!

PR #21, opened 2015-02-16
Visual Changes 3

fluddokt:
Combo Color ordering 
General Image scaling
Score fixed size width
Hard coded Ranking Panel
Hard coded Scorebar
Made MenuButtonBG colours darker
Extends ScoresData bg to clock

Spinner centering (slick has bad centering)


itdelatrisu:
Looks great so far, will look more thoroughly later when I have time.  Things I noticed: the "skip" button is huge (the default one, at least, which should be the same size as the back button?).  Can you also merge my recent commits so I can merge the pull request automatically?  Thanks!

fluddokt:
I just tested it the res folder as a skin in osu, and it really is that big in osu. Also can I remove the 4-5 second minimum for the skip button to be displayed at all. I'll try to get to merging in a few hours.

PR #25, opened 2015-02-27
Allows multiple instance of the same clip to be played simultaneously.

fluddokt:
This might fix some issues with some sounds not being played sometimes.

Also might require more memory since more clips are created.

### DIFF
--- a/src/itdelatrisu/opsu/audio/HitSound.java
+++ b/src/itdelatrisu/opsu/audio/HitSound.java
@@ -20,8 +20,6 @@ package itdelatrisu.opsu.audio;
 
 import java.util.HashMap;
 
-import javax.sound.sampled.Clip;
-
 /**
  * Hit sounds.
  */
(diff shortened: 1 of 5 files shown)


itdelatrisu:
Could the extra clips be destroyed afterwards?  Just from one play of [this map](https://osu.ppy.sh/p/beatmap?b=131891), about 40 clips get created and my memory usage went up by ~80mb with the default skin.  That's probably too much...

fluddokt:
Yea, but it becomes a question of when to destroy it. 
It could be destroyed right after each clip is done playing, but opening and closing clips is bad I think.
After playing the beatmap?
But if its reaching that much clips, then its probably going to need about that much clips to play another beatmap.

itdelatrisu:
Yeah, I can't think of a good solution.  The game hangs a lot when closing clips, so closing them immediately isn't an option.  But it seems silly to allocate so much extra space for sound effects...

How does this sound?  We could set a reasonable global limit on the number of Clips created, and whenever that limit is reached, just reuse the last clip played from each MultiClip (e.g. cutting off the original).  We could close all the extra clips before/after playing a beatmap (maybe in a separate thread?).

fluddokt:
Maybe have a limit on each MultiClip instead, so that one MultiClip can't just steal all the extra clips?

itdelatrisu:
I think it's actually okay as-is, since extra space gets allocated to whatever clips need them most.  I'll merge this now and make a couple of small changes (reducing number of methods, setting MAX_CLIP to 20, etc.).  Thanks so much!

PR #28, opened 2015-03-06
Attempt to fix jitter by using a higher precision timer.

itdelatrisu:


chong601:
The jitter is reduced a bit but not completely D:
Most noticeable at sliders

PR #31, opened 2015-03-07
Remove restriction from left/right click keys

ScoreUnder:
I use the comma and period keys to click with. Since I have the tablet directly in front of my keyboard, that stops me from having to reach all the way over to the z/x keys which would be quite uncomfortable.

There's no reason why someone can't have ctrl/alt or home/pgup as their preferred keys either. At the moment I think just escape needs to be checked for to make this "idiot proof", but I would question the usefulness of idiot-proofing to that degree. I've done it anyway.


itdelatrisu:
Looks fine.  Thank you!

PR #37, opened 2015-03-07
Spinner Changes

fluddokt:
frame independent spin up/average
rpm for Auto/SpunOut
CCW rotation
fade in
addition


itdelatrisu:
Wow, thanks so much!  The only thing I noticed was that the "spun out" speed was slightly off (had to change divisor from `32` to `33.25`), but this looks perfect otherwise.  Thanks as always!

PR #38, opened 2015-03-07
Next track on trackEnded

fluddokt:
(but checks trackended every update...)
Continue playing while minimized(does not fix moving)
start from preview time in songMenu on repeat


itdelatrisu:
Thanks!  I was trying to figure out a way to do this without checking on every update(), but whatever.  I'll follow-up shortly by looping tracks in the other states.

Issue #33, opened 2015-03-07
Could not play track


Lucki:
I've a problem with this Beatmap, all others seems fine so far, or I haven't found one with similar error.
When clicking on it, a error-window pops in front containing this:
~~~
Could not play track 'Nightcore - Where Is The Love.mp3'.
java.lang.RuntimeException: Resource not found: /home/lucki/.local/share/opsu/Songs/19882 Baracuda - Where Is The Love (Nightcore Mix)/Nightcore - Where Is The Love.mp3
	at org.newdawn.slick.util.ResourceLoader.getResourceAsStream(ResourceLoader.java:69)
	at org.newdawn.slick.openal.OpenALStreamPlayer.initStreams(OpenALStreamPlayer.java:149)
	at org.newdawn.slick.openal.OpenALStreamPlayer.play(OpenALStreamPlayer.java:196)
	at org.newdawn.slick.openal.StreamSound.playAsMusic(StreamSound.java:46)
	at org.newdawn.slick.Music.startMusic(Music.java:314)
	at org.newdawn.slick.Music.loop(Music.java:291)
	at org.newdawn.slick.Music.loop(Music.java:264)
	at itdelatrisu.opsu.audio.MusicController.playAt(MusicController.java:136)
	at itdelatrisu.opsu.audio.MusicController.loadTrack(MusicController.java:119)
	at itdelatrisu.opsu.audio.MusicController.access$000(MusicController.java:42)
	at itdelatrisu.opsu.audio.MusicController$1.run(MusicController.java:86)
Sat Mar 07 13:37:32 CET 2015 WARN:OpenALStreamPlayer: setPosition: failed to skip.
~~~

Edit: Holding F2 gives me a second and third Song with error:
~~~
Could not play track 'Electric Six - Gay Bar.mp3'.
java.lang.RuntimeException: Resource not found: /home/lucki/.local/share/opsu/Songs/9550 Electric Six - Gay Bar/Electric Six - Gay Bar.mp3
	at org.newdawn.slick.util.ResourceLoader.getResourceAsStream(ResourceLoader.java:69)
	at org.newdawn.slick.openal.OpenALStreamPlayer.initStreams(OpenALStreamPlayer.java:149)
	at org.newdawn.slick.openal.OpenALStreamPlayer.play(OpenALStreamPlayer.java:196)
	at org.newdawn.slick.openal.StreamSound.playAsMusic(StreamSound.java:46)
[...]

PR #39, opened 2015-03-08
Fixes some sliders with too much control points. / Attempts to load music with incorrect file names. #33

fluddokt:
Fixes some sliders with too much control points.
Attempts to load music with incorrect file names. #33


itdelatrisu:
Great, thanks so much!

PR #40, opened 2015-03-08
Allow unlimited FPS to partially fix the jitter issue

chong601:
Added 500, 1000 and unlimited fps
Fixes:
Jitter during gameplay (still occurs on really fast maps)

Issues found (for now):
Spinners are unclearable with fps>240.
Combobursts moves in faster than normal
Cursortrail doesn't render properly with fps>240


itdelatrisu:
So, do the GameContainer changes actually fix anything?  (I can't see any difference on my computer.)

I don't really want to add any more FPS settings, since:
* I don't think opsu! would benefit at all from >240fps (which, in my opinion, is already really high).  I haven't played any games besides osu! that allow this, either.
* Just running a quick search, it seems like osu! has plenty of issues with high FPS with OpenGL.
* I won't be able to resolve any issues at higher FPS settings because my computer won't run at much higher than 240fps.

chong601:
Well, actually at 240 fps, the game is actually pretty playable at the moment.
But in the case of my PC, i need about >500 to actually somehow remove the jitter movement (if not completely, but at least mostly) :(.

![My opsu! fps] (http://puu.sh/gsgbl.png)

Tested on some of my friends PC and some even reach >12k fps with quad GTX 980

(but I do really need to look at the spinner since it is really really broken at high fps lol)

itdelatrisu:
I still don't understand why having outlandishly high FPS is necessary, though.  I think pushing the FPS is the wrong approach to fixing any jitter you're experiencing...

chong601:
I do agree on your explaination but for now with the current game rendering (which uses updateAndRender() call), the only workaround for the jittery movement is by ramping the FPS to the max (or arbitrary values above the 240 fps value).

The probable reason is because the update() call (which handles the game logic) is slower than render() call (which draws the game to the screen). From the current code, the game will update **and** render the game logic at the same time (and this maybe related to my current graphic card can't do OpenGL stuff as good as DirectX).

I am planning to rewrite the rendering part to try fix the jitter part by separating the updateAndRender() calls to two different call (mainly one update() and one render() call) to run the update() (which handles the game logic) more frequently rather than render() which puts the last update() call to the screen (after looking at some other games using Slick2D where they run update() twice the times of render() call for example 120fps to refresh the game logic and only use 60 fps to display it to the screen).

I might start do the fix on April (after my internship ended) and will try to fix the jitter.

chong601:
Closing this pull request (for now) since i have no ETA on when I can put a fix :(

PR #49, opened 2015-03-14
@2x size images.

fluddokt:
ComboBurst speed scaling.

CircumscribedCircle drawing optimization.


itdelatrisu:
Great, thanks!  Nice idea to just call `getScaledCopy(0.5f)`, I didn't think of that...

Issue #50, opened 2015-03-15
[enhancement] Reverse the sliderball values rather than turning the ball direction


chong601:
On sliders with reverse, the sliderball is set to turn 180 degrees when they return.
On the opsu! implementation:
Before reverse:
b0=>b1=>b2...=>bX where X is the largest value of the sliderb* number.
After reverse
But when it reverses, it uses the bX<=...b2<=b1<=b0 and the ball is turned back to the direction of the starting position.

The original osu! implementation uses this direction:
Before reverse:
b0=>b1=>b2...=>bX
After reverse:
b0<=b1<=b2...<=bX

chong601:
Confirmed it worked as expected

PR #52, opened 2015-03-18
Replay Changes

fluddokt:
Saves replay frames as if its repaying.
Then plays the replay at exact track position.
This should give out exact results as the score screen.

Also Calculates Md5 for beatmap files. (doesn't currently help anything)
and Reverse Slider ball #50


itdelatrisu:
The replays look good, but from some quick tests, this breaks a few things:

* The "autopilot" mod (recently implemented) is broken during replays -- it doesn't record the user's clicks.
* Loading checkpoints sometimes causes instant death.
* The pulsing cursor animation when returning from the pause screen is broken (no longer animated).
* The "skip" button isn't processing hover effects during lead-in time.

I don't know if anything else is broken, but could you do a bit more testing?  (And comments on all the new methods would be helpful, too!)  Thanks!

fluddokt:
Want to split the Game class into two classes?
Such that there is a Game State and a ?Active? Game.
So that we can maybe implement ghost replay/ multi replay?

itdelatrisu:
Awesome, it looks good now -- I'll make a few changes as usual, but let me know if I do anything you don't like.

I do think it might be a good idea to create a separate Game class, though it's not really my priority right now.  I'm not sure how performance will be if we try to run multiple replays at once... and I also don't want to write a UI for that. XD  What do you mean by "ghost replay"?

fluddokt:
I don't think the performance will be too bad, if they are all drawn in the same view (just the cursors and maybe hit results).
Though I do also want to be able to display it as separate views (Like 4 views on the screen).

Yeah, writing a UI for this is going to be annoying, I'm not even sure how you will go about selecting the replays you want.

Oops, I meant [Play against Ghost](http://osu.ppy.sh/forum/viewtopic.php?f=4&t=11114)
Which is like playing against you replays as if you were playing multiplayer.
Which will display the scores (and maybe the cursor) of the replay at your current position during play.

I'm also thinking about adding some ways to import osu replays, so that we can compare the difference in scores and try to make it closer to osu (and also for mutli replay if it ever gets implemented).

itdelatrisu:
I see -- thanks for explaining.  The "ghost" thing seems pretty cool, and I don't think it'd be difficult to implement.  As for the UI, a cop-out solution could be to just put the new features into new `ButtonMenu` states, maybe branching off of the existing `SCORE` (score management) state.  Not a very intuitive location, but it'd save a lot of trouble. XD

Importing replays should be very straightfoward -- just parse the file and stick it in the score database (with whatever the current file name is).  I can't think of any way of verifying that replay files actually belong to a beatmap, though (I can't find information on osu!'s hashing schemes), and I also can't think of any way to select the beatmaps other than with Swing (which is really ugly).  LWJGL3 added drag-and-drop support, but I highly doubt anyone will port Slick2D to LWJGL3.

I'm probably not going to have much time to work on this for at least a few weeks, since I really need to start working on school assignments soon.  x.x  Feel free to leave me any messages though, and I'll try to respond asap.

fluddokt:
Yeah that could probably work. I'll try that if and when I ever get around to this.

I'm currently thinking of just doing what it does with importing osz files. Just stick some replays in some folder and it will be imported on launch or something. The beatmap hash in the replay file refers to the md5 hash of the beatmap file (which I got from [osu wiki](https://osu.ppy.sh/wiki/Osr_%28file_format%29)), so we can just use that to figure out which beatmap the replay refers to. So will probably need to pre-calculate this hash for each beatmap and store it somewhere. I've already added a method to calcuate the md5 hash of a file in Utils in this pull request.

PR #63, opened 2015-03-30
Circle and slider fade in.

DarkTigrus:
Match osu! behaviour. Fadein time is static and doesn't scale with approach rate. Circles, numbers, arrows and ticks alpha is the same.

### DIFF
--- a/src/itdelatrisu/opsu/GameData.java
+++ b/src/itdelatrisu/opsu/GameData.java
@@ -444,13 +444,13 @@ public class GameData {
 	 * @param y the center y coordinate
 	 * @param scale the scale to apply
 	 */
-	public void drawSymbolNumber(int n, float x, float y, float scale) {
+	public void drawSymbolNumber(int n, float x, float y, float scale, Color color) {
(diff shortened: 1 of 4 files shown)


itdelatrisu:
The static fade in time looks fine -- just wondering, though, where did you get that information?  And is 375ms just a guess, or an actual value?

I really don't like the fading numbers -- I can't really describe why, but it feels uncomfortable.  The other things seem okay.  Thoughts?

DarkTigrus:
It's a guess. I made custom beatmap, skin, recorded it on video and it timed frame by frame. Checked approach rate 1, 5, 10. Got various results between 364ms and 387ms due to video framerate(?). ~375ms is average.

Static 1f alpha on numbers looks really wierd to me. It is not how in osu! Also looks ugly at streams.

itdelatrisu:
Okay, that makes sense -- thanks for the screenshot!  I'll change the image drawing to set an alpha level instead of drawing with a color filter, but everything looks good otherwise.

PR #68, opened 2015-03-31
HalfTime and DoubleTime mod

DarkTigrus:
Dynamic bpm, map length in song select menu.
Fixed replay and spinner.


itdelatrisu:
Thank!  Is there a way to make it so that the actual pitch of the song isn't changed, though?  That was the main difficulty with the speed-changing mods. x.x

What are the "replay and spinner" fixes?

DarkTigrus:
Seems like Slick and LWJGL in general don't have any way to change audio tempo.
Spinner relies on ```update()``` ticks and not ```trackPosition```. Not really a fix, but now should work with DoubleTime.
I moved ```MusicController``` below replay initialization, so DoubleTime should work properly with replays.

DarkTigrus:
Some updates:

  * Added ability to change pitch on fly. I tested it only on Linux. It should be tested on other systems because changing gain is not working on my OS. Really weird.
  * Added playback speed in replays.
  * AutoMod counts as replay.

I used different font for buttons because I couldn't import kochi-gothic one. Probably you want to redraw them.

itdelatrisu:
Eh, I'm still kind of undecided about this.  I just can't imagine anyone using the DT/HT mods if we can't get this to work without pitch shifting (but I couldn't find any implementations of this in OpenAL...).  Thoughts?

Anyhow, comments:
* When watching a HT replay, you should be able to play back at 0.5x speed (currently, only 1x and 2x are available; this is different from DT).
* When leaving the game state, the speed multiplier should be reset.
* `GameMod.justChanged()` feels really hackish, and it'd be fine if you just reset `songInfo[]` when entering the song menu.
* I don't think you should be able to change the `Image` in a `MenuButton`... just initialize all three buttons and draw whichever is active.
* I think the playback buttons look okay, but maybe they should fade in (or change colors, like in osu?) instead of expand when hovering.

DarkTigrus:
I did some quick research about openal before. Nothing useful.
Well, if you want to bring up advanced osu players you want to have DT even if it sound like nightcore. Personally i don't mind it . But i see why people can be annoyed.

itdelatrisu:
Okay, I'll just merge this then; I suppose it doesn't hurt.  Thanks so much!

PR #71, opened 2015-04-06
Added hit animations, updated lighting & added InteliJ to gitignore

LemonLake:
I hope I did this right ._. pardon the IntelliJ typo


itdelatrisu:
Thanks for taking care of this!  There are a couple of bugs:
* The hit circles are being drawn for slider tick results.
* In osu!, the first and last circles in a slider get animated after completing the slider; in your implementation, the first circle is animated immediately.
osu! also has a kind of bouncy effect when you hit a circle, and it feels a little off seeing just a linear size increase.

There's a copyright issue, as well -- peppy has copyright on all the default osu! skin images, so I can't actually include the `lighting.png` image in this repository.  Users will have to supply it themselves.

LemonLake:
This is the lighting texture I use in my own skin, so this one should be fine. I'll get to work on those bugs right away

LemonLake:
I'm not quite sure what you mean by bounciness, but it definitely does look like there is some easing (60fps at 30% speed)

Bigpet:
@LemonLake just implement whatever you need on top of the current slider system (something almost exactly like it will still exist as a fallback anyway), I'll rebase however necessary.

LemonLake:
New bug: Geki and katu hit results no longer appear. Know the solution, just looking for the fix

src/itdelatrisu/opsu/objects/HitResultType.java
```diff
@@ -0,0 +1,10 @@
+package itdelatrisu.opsu.objects;
+
+public enum HitResultType {
+    CIRCLE,
+    SLIDERSTART,
```
itdelatrisu: `SLIDERSTART` is never used.

src/itdelatrisu/opsu/GameData.java
```diff
@@ -1152,11 +1244,12 @@ public void hitResult(int time, int result, float x, float y, Color color,
 		case HIT_MISS:
 			hitValue = 0;
 			changeHealth(-10f);
+			changeHealth(-10f);
```
itdelatrisu: Please explain?

src/itdelatrisu/opsu/Utils.java
```diff
@@ -607,4 +607,14 @@ else if (seconds < 3600)
 		else
 			return String.format("%02d:%02d:%02d", seconds / 3600, (seconds / 60) % 60, seconds % 60);
 	}
+
```
itdelatrisu: Could you add Javadocs for these?

itdelatrisu:
This looks really great now -- thanks again!  I added a few comments in the diff, but visually this seems perfect.  Are you ready for me to merge it?

src/itdelatrisu/opsu/objects/HitResultType.java
LemonLake: That was going to be used, but I never actually used it. I left it in there in the event that anyone in the future wanted to use it.

src/itdelatrisu/opsu/GameData.java
LemonLake: I have no idea how that happened.

src/itdelatrisu/opsu/Utils.java
LemonLake: Will do.

LemonLake:
I'll just fix those and then it will be ready!

LemonLake:
Go ahead and merge!

PR #75, opened 2015-04-12
Minor changes to opsu! UI

chong601:
Set the main menu top and bottom bar alpha level to 50%
Set the song select top and bottom bar alpha level to 100%
Slight changes to the wording on main menu


itdelatrisu:
Eh... I don't really like how these changes look (or maybe I'm just used to the current appearance).  I don't think the main menu changes are necessary, and I had changed the song menu bar transparency a while ago but neither I nor fluddokt liked it.

PR #76, opened 2015-04-12
More curve types

fluddokt:


itdelatrisu:
Awesome, thank you!! It took me a while to even find maps with Catmulls, but these look wonderful.  Again, thanks so much for helping out!

itdelatrisu:
Actually, in the Easy mode of [this map](https://osu.ppy.sh/s/302), the Catmulls at the beginning are totally wrong... could you have another look?

PR #97, opened 2015-06-21
Error handling and cleanup for options

wacossusca34:


src/itdelatrisu/opsu/Options.java
```diff
@@ -223,10 +225,13 @@ public void click(GameContainer container) {
...
 			public void read(String s) {
+				String str = "?";
 				try {
-					Resolution res = Resolution.valueOf(String.format("RES_%s", s.replace('x', '_')));
-					resolution = res;
-				} catch (IllegalArgumentException e) {}
+					str = String.format("RES_%s", s.replace('x', '_'));
+					resolution = Resolution.valueOf(str);
+				} catch (IllegalArgumentException e) {
+					ErrorHandler.error(String.format("Invalid resolution in config file: %s", str), e, false);
```
itdelatrisu: All other errors in reading the configuration file are ignored, which is why I purposefully didn't throw an error for this, either.

src/itdelatrisu/opsu/Options.java
```diff
@@ -1065,12 +1070,16 @@ public static File getBeatmapDir() {
 			return beatmapDir;
 
 		// search for directory
-		for (int i = 0; i < BEATMAP_DIRS.length; i++) {
-			beatmapDir = new File(BEATMAP_DIRS[i]);
+		for (String BEATMAP_DIR : BEATMAP_DIRS) {
```
itdelatrisu: This change isn't really necessary... and `BEATMAP_DIR` shouldn't be in all caps imo.

src/itdelatrisu/opsu/Options.java
```diff
@@ -1065,12 +1070,16 @@ public static File getBeatmapDir() {
...
+			beatmapDir = new File(BEATMAP_DIR);
 			if (beatmapDir.isDirectory())
 				return beatmapDir;
 		}
-		beatmapDir.mkdir();  // none found, create new directory
+		// none found, create new directory
+		assert beatmapDir != null;
```
itdelatrisu: I'd rather not include asserts.  If something goes wrong, let an error get thrown.

src/itdelatrisu/opsu/Options.java
```diff
@@ -1229,7 +1244,7 @@ public static void parseOptions() {
 
 		// create option map
 		if (optionMap == null) {
-			optionMap = new HashMap<String, GameOption>();
+			optionMap = new HashMap<>();
```
itdelatrisu: I prefer the former for the sake of clarity (and consistency).

itdelatrisu:
I've left comments in the diff, but in summary: the errors in directory creation are fine, but I don't think any of the stylistic changes are necessary.  I can make my own commit with the error handling, or do you want to follow up with the pull request?

wacossusca34:
The stylistic changes are my IDE. I need to turn those off, this was a bit of a premature PR anyway.

Lots of... bad error handling and resource management in opsu though. The `System.gc()` calls to handle out-of-scope IO streams was definitely interesting to come across.

itdelatrisu:
The `System.gc()` calls aren't necessary, but it does help keep memory usage much lower than it'd be otherwise (at least on my machine).  I actually can't remember what the issue was exactly, but I think audio/graphics resources weren't getting collected nearly frequently enough, whereas they really should be after every game.  There might be a better solution, so let me know if you have any ideas.

wacossusca34:
You shouldn't really need to call `System.gc()`. If you want the app to use less memory, just set a smaller maximum heap size when you are running it. If memory usage is over the minimum heap size, it will collect out-of-scope memory when the VM feels it needs to, and will ensure that it will not hit the maximum heap size when there is out-of-scope (garbage) memory that can be collected. 

The two instances in `Game.resetGameData()` and `MusicController.play(...)` are unnecessary, but not necessarily problematic. I was referring to this:

    System.gc();  // TODO: why can't files be deleted without calling this?

In `BeatMapSetList.deleteSongGroup(...)`. If you cannot delete files without calling System.gc, you have descriptors/streams that haven't been closed for the same files (and therefore the Java VM is closing them on the `finalize()` methods).

itdelatrisu:
I don't think the JVM GC options are sufficient, though -- limiting the maximum heap size isn't what we want, and if we just lower the threshold for when collection takes place, it could happen when we don't want it to (e.g. during gameplay).

I'd forgotten about the instance in `deleteSongGroup()`, though.  If I recall correctly, it was specifically the audio streams that weren't getting closed, but nothing I tried could fix the issue.  Any help there would be appreciated.

wacossusca34:
> I don't think the JVM GC options are sufficient, though

It's how Java works. If you want to have more interaction with memory, you have to go native. The Java spec and VM is meant to provide a large amount of abstraction for memory, to the point in which the way the VM (and its garbage collector) is implemented is really only dictated by the bytecode format and java language features.

And `System.gc()` isn't even guaranteed to do anything.

That being said, most GC implementations are very smart and can work with memory very quickly without affecting performance of the application, unless you are churning out/processing massive amounts of data (gigabytes).

> it could happen when we don't want it to (e.g. during gameplay)

The only time it could affect gameplay is if the maximum heap size is being reached (where the infamous stop-the-world GC cycles can run), otherwise garbage memory will silently be collected without halting anything (and having extra garbage memory in the heap won't affect the performance of the app).

> I'd forgotten about the instance in deleteSongGroup(), though. If I recall correctly, it was specifically the audio streams that weren't getting closed

I can try finding out what went wrong. I just picked up the repository a few hours ago and I'm still exploring the code.

Sorry if I'm being nitpicky or critical, this is actually an impressive project and I'm having fun finally playing the game smoothly on Linux :)

Issue #67, opened 2015-03-31
Spinners gets harder to clear if the fps is higher


chong601:
On 60 fps, the spinner is almost osu! like clear.
But on 240 fps, the spinner is really hard to clear (especially on short spinners)

I also have added the replays since it appears to be affecting replays too
The map played is [Primastella - Koigokoro [iyasine's Hard]] (https://osu.ppy.sh/b/326600)
[First play with 120 fps] (http://puu.sh/gWmM1.osr)
[Second play with 240 fps] (http://puu.sh/gWmMl.osr)

On 1st play, I manage to get a Clear with 2000 score (idk how to call it)
But on 2nd play, I only manage to get a 100 (not even a clear)

However, showing the replays on different fps (120 fps replay on 240 fps or 240 fps replay at 120 fps) doesn't affect the spinner in any way.

itdelatrisu:
I thought this was fixed in #37. x_x
Anyway, kind of busy now, but will take a look when I can.

chong601:
it does fixed (since i am mostly on 120 fps) but when i accidentally switched to 240 fps, this happens

fluddokt:
240fps
AngleDiff 0.0
AngleDiff 0.39490795
AngleDiff 0.0
AngleDiff 0.40509343
AngleDiff 0.34233928
AngleDiff 0.0
AngleDiff 0.28742528
AngleDiff 0.0
AngleDiff 0.24742842
AngleDiff 0.0
AngleDiff 0.24211335
AngleDiff 0.0
....

120fps
AngleDiff 0.24578154
AngleDiff 0.24661213
AngleDiff 0.25922775
AngleDiff 0.2885722
AngleDiff 0.34786296
AngleDiff 0.34788835
AngleDiff 0.35399747
AngleDiff 0.32616687
....


So it seems you pointing device(tablet?) updates at around 120hz
and the spinner is based on the average instantaneous angular velocity
so your rpm is pretty much halved.

I'll try to think of something to fix it.

FPS for Replay doesn't affect it since it actually plays its back at the replays FPS even if your FPS is different.

chong601:
i am using mouse on that replay (tablet is affected as well)

1338h4x:
Spinners constantly get stuck for me on Android, even when locked at 60FPS. I have to turn on Spunout to deal with them, and even then I can see them visually getting stuck (but thankfully they still clear).

itdelatrisu:
[...]

PR #99, opened 2015-06-22
Replay Importing / Spinner

fluddokt:
Imports replays/.osr from the ReplayImport directory into opsu 

Slightly improved score accuracy.
Improves Spinner and should fix? #67
 but still probably wrong.

Poorly implemented replay seeking. Clicking near the top of the window seeks the position of the replay
where the left is the start and the right is the end.


src/itdelatrisu/opsu/GameData.java
```diff
@@ -296,10 +296,6 @@ public HitObjectResult(int time, int result, float x, float y, Color color,
 	/** The difficulty multiplier used in the score formula. */
 	private int difficultyMultiplier = 2;
 
-	/** Beatmap HPDrainRate value. (0:easy ~ 10:hard) */
-	@SuppressWarnings("unused")
-	private float drainRate = 5f;
```
itdelatrisu: This is going to be used in the health formula eventually, so please leave it there!

src/itdelatrisu/opsu/beatmap/BeatmapParser.java
```diff
@@ -589,6 +589,8 @@ else if ((type & HitObject.TYPE_SLIDER) > 0)
 		// parse hit objects now?
 		if (parseObjects)
 			parseHitObjects(beatmap);
+		
+		beatmap.md5Hash = Utils.getMD5(file);
```
itdelatrisu: There should be a way to calculate the MD5 without reading the file twice -- we should definitely do that instead, since this will probably hurt loading times.

itdelatrisu:
Sorry it took so long to merge this.  Thanks again!

I'll try to come up with a better UI for the replay seeking, but I can't think of anything at the moment. x.x

Issue #101, opened 2015-06-24
Fullscreen shrinks when objects appear


LemonLake:
I think this happens when sliders appear, probably due to FBOs (if they're even used, i have no clue)

Video of the issue:
https://www.youtube.com/watch?v=uo2YD8NuWDc

Bigpet:
How do you set it to fullscreen other than setting it to the resolution of your monitor? Also, what's your GPU?

LemonLake:
I just set it to the resolution of my monitor. Nvidia GeForce GTX 760.

Bigpet:
Is there like a method that gets called in a hierachy when things shut down? Sort of like "init" but for shutdown? Because there's some static state that I should probably clean up if the process isn't going to be killed when the OpenGL context gets killed.

itdelatrisu:
Most of the cleanup happens in `Container.close_sub()`,

PR #102, opened 2015-06-24
Fullscreen resizing fix 

Bigpet:
related to #101

Correctly clean up the created OpenGL objects created for the slider rendering.

This is necessary because the OpenGL context being closed does apparently not close
the process and the game can be restarted with another resolution without shutting
down the process completely.

Not really sure what else to do to cleanly shut this down, I'll add some more comments in a moment


Bigpet:
I think that should be it.

I never got the error without switching the resolution with "Ctrl+Shift+F5", so if that was an issue I couldn't reproduce it.

itdelatrisu:
Thanks for taking care of this!

PR #108, opened 2015-07-08
Backup and restore the viewport size when rendering sliders

### LINKED ISSUE #107: Flashlight rendering problems with new slider style
When enabling the "flashlight" mod with the new slider style, this happens:
ss

@Bigpet: Could you take a look at this when you get a chance?

Bigpet:
Backup and restore the viewport size when rendering sliders. Needed because Slick tends to allocate offscreen buffers for itself only with power of two textures, so it will use another viewport when rendering to its own offscreen buffers.

Should fix #107

### DIFF
--- a/src/itdelatrisu/opsu/render/CurveRenderState.java
+++ b/src/itdelatrisu/opsu/render/CurveRenderState.java
@@ -27,6 +27,7 @@ import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 
 import org.lwjgl.BufferUtils;
+import org.lwjgl.opengl.EXTFramebufferObject;
 import org.lwjgl.opengl.GL11;
 import org.lwjgl.opengl.GL13;
 import org.lwjgl.opengl.GL14;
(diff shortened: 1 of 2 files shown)


Bigpet:
uh, wait one sec, there's a small error

Bigpet:
Alright, that should be it. I should really always do a complete diff before making a pull-request

itdelatrisu:
Thanks!  It doesn't seem to completely fix the issue for me, though:
ss

Bigpet:
what's the beatmap?

itdelatrisu:
[221216](https://osu.ppy.sh/s/221216) (hard)

Bigpet:
I'm pretty sure you told me like twice already but what was your GPU again? Just asking because this doesn't happen with my Nvidia gpu, gonna try on my intel gpu next

itdelatrisu:
Playing at 1440x900, if that matters.

Bigpet:
Alright, I can reproduce it with my Intel card now. I'll take a look at it tomorrow

itdelatrisu:
Awesome, thank you!

Bigpet:
Turns out the Intel drivers return the same FBO id twice if the ARB and EXT versions of framebuffers are used. And the flashlight mod calls Slick methods that use the EXT version, so I needed to use those (the only other alternatives would have been some ugly hacks).

This should remedy the situation and fix #107 in its entirety.

itdelatrisu:
Perfect.  Thanks again!

PR #115, opened 2015-08-08
'Hidden' gamemod

mad-s:
Enables the 'hidden' game-mod and changes circles and sliders accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 .metadata
 .classpath
 .project
+.externalToolBuilders/
 
 # IntelliJ
 .idea/
--- a/src/itdelatrisu/opsu/GameMod.java
+++ b/src/itdelatrisu/opsu/GameMod.java
@@ -48,7 +48,7 @@ public enum GameMod {
 	              "DoubleTime", "Zoooooooooom."),
 //	NIGHTCORE     (Category.HARD, 2, GameImage.MOD_NIGHTCORE, "NT", 64, Input.KEY_D, 1.12f,
 //	              "Nightcore", "uguuuuuuuu"),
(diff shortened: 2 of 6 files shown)


itdelatrisu:
Awesome, thanks so much!  I'm glad the mods are finally finished now (well, the basic ones, at least) :D!

This was my first time actually trying the mod in osu!, so let me know if I'm mistaken about anything.  That said, here are differences I noticed from osu!:
* The slider follow circle should still appear with the mod enabled.
* The `SPINNER_CIRCLE` element should not appear with the mod enabled.

Lastly, some questions about `decayTime`:
* Is 800 a value you estimated, or is it the actual value?  It seems pretty similar to osu!, but if it's not an exact number, just add a TODO to make that clear; otherwise, what's your source?
* Does the value ever change (with difficulty settings, etc.)?  If not, it can just be a public static constant, and doesn't need an instance accessor.

mad-s:
I re-enabled the slider follow circle and disabled the spinner circle. The `decayTime` value is just an estimate. My only source of information are videos (I run Linux) and they are not very consistent. Maybe someone can measure the decay time. 
Does the spinning meter appear in hidden mode?

abalabahaha:
Just tried Ievan Pollka, apparently the spinner completion meter still appears

itdelatrisu:
Great, thanks for the fast reply.  I'll go ahead and merge this now, but feel free to keep commenting here and/or submit another pull request if you want to add anything.

The spinner meter is still present,but the spinner approach circle should also be hidden -- sorry I didn't catch that earlier.  I'll change that one myself.

itdelatrisu:
Wow, I'm really sorry, I provided a lot of wrong information... it's not the spinner circle that disappears, but only the spinner approach circle.  I totally forgot what my skin looked like.  I'll fix that as well.

Issue #120, opened 2015-08-13
Slider bodies don't show up at all on 0.10.1 on linux 64 (nouveau, GeForce GT 640)


lollllloro:
![2015-08-13 01 24 23_1440x900_scrot](https://cloud.githubusercontent.com/assets/13772620/9240607/d7c98144-4170-11e5-9802-1814e8c5260e.png)

**The issue is NOT present on 0.10.0**

tested with 64bit:
openjdk 1.8.0_66
openjdk 1.7.0_79
oracle java 1.8.0_51

Also, this cursor: ![cursor](https://cloud.githubusercontent.com/assets/13772620/9240721/39ab41d0-4172-11e5-9794-487d85a9e430.png) doesn't look like itself at all on any version of opsu I've tried on linux, it's always like [this](https://cloud.githubusercontent.com/assets/13772620/9240752/c2bf6a14-4172-11e5-8179-2cabf09d0d41.png) (there are no artifacts in the image, it always has that square around it)
This is tested on multiple computers, including 32b ones.

A minor thing I noticed too: if a beatmap with ogg music named *.mp3 (osu plays it normally) is selected, opsu loses sound and gets very laggy, java stays 100% usage on a single core and a "Failed to destroy OpenAL" exception gets thrown every time opsu tries to load another song. This is also tested with the versions above.

.opsu.log after doing all the tests (gzipped 5415 B):
http://54.213.154.174:777/B/.opsu.log%202015-08-13%2002:45:45.gz
Opsu crashes in the end when I move the beatmap dir in question away from Songs while running with oracle java.

itdelatrisu:
Thanks for the detailed report.  I'm unfortunately not familiar enough with some of these issues to provide solutions, but maybe someone else can help.

* Is the slider issue still present if you revert #108?  @Bigpet, do you have any insights on this?
* I don't have any ideas about the why the cursor isn't rendering correctly. x.x
* Again, not really sure what the problem is, but I think this is the relevant piece of code (written by fluddokt) in the modified `org.newdawn.slick.openal.OpenALStreamPlayer.java`:
```
			if (ref.toLowerCase().endsWith(".mp3"))
				audio = new Mp3InputStream(ResourceLoader.getResourceAsStream(ref));
			else
				audio = new OggInputStream(ResourceLoader.getResourceAsStream(ref));

			if (audio.getRate() == 0 && audio.getChannels() == 0) {
				if (ref.toLowerCase().endsWith(".mp3"))
[...]

PR #122, opened 2015-08-13
Lower GLSL version requirement

Bigpet:
I didn't really use any of the newer features anyway.
This should accomodate open source drivers that use older versions of MESA.

Related to #120

### DIFF
--- a/src/itdelatrisu/opsu/objects/curves/Curve.java
+++ b/src/itdelatrisu/opsu/objects/curves/Curve.java
@@ -87,12 +87,12 @@ public abstract class Curve {
 		Curve.borderColor = borderColor;
 
 		ContextCapabilities capabilities = GLContext.getCapabilities();
-		mmsliderSupported = capabilities.GL_EXT_framebuffer_object && capabilities.OpenGL33;
+		mmsliderSupported = capabilities.GL_EXT_framebuffer_object && capabilities.OpenGL30;
 		if (mmsliderSupported)
 			CurveRenderState.init(width, height, circleSize);
 		else {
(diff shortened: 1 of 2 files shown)


mad-s:
Please lower the OpenGL version requirement in src/itdelatrisu/opsu/objects/curves/Curve.java:90 to OpenGL 3.0 instead of OpenGL 3.3

Bigpet:
@mad-s I guess that would work fine. Technically we would need to check if glsl version "130" is supported too. But I don't know of a driver that supports GL 3.0 but not GLSL 130

mad-s:
According to https://en.wikipedia.org/wiki/OpenGL_Shading_Language#Versions OpenGL 3.3 corresponds to GLSL 3.3 and OpenGL 3.0 corresponds to GLSL 1.3. This is also the reason for 940e9baa41674d5a272bfad2a55f53a86d518b51, which fixed slider rendering for my GPU.

Bigpet:
@mad-s well that means that GLSL 330 and GL 3.3 were released alongside. But if you read the OpenGL specification then you'll realize that those standards are still seperate and to get a conforming OpenGL 3.3 implementation you do not actually need to support GLSL 330. (If this were the case then #120 would be a bug in noveau, but it's not)

itdelatrisu:
Thanks again for taking care of this.  Though, why aren't we checking if GLSL 130 is supported?

Bigpet:
@itdelatrisu because I'm lazy :smile: . Also, there's no really good way to poll it other than writing a minimally viable shader program in that version and try to compile and link it. You can only query the highest supported version and I don't think an implementation has to support all previous versions.

PR #127, opened 2015-08-27
Added cursor animation

LemonLake:


itdelatrisu:
Looks good in-game, the only thing that kind of caught my attention was this:
```
			cursorSizeAnimated = (mousePressed ? 1f : 1.25f) +
					((mousePressed ? CURSOR_SCALE_CHANGE : -CURSOR_SCALE_CHANGE) * ...
```
I think this would be a lot clearer as just an `if (mousePressed) ; else ;` but I'll probably merge it later anyway if you don't feel like changing this.

Issue #133, opened 2015-09-16
Snaking sliders


FenekAlfa:
"Snaking sliders" are sliders that render gradually from start to end instead of instantly.

I didn't really notice it when Hidden wasn't available, but with HD I mess up sliders all the time because I'm too used to them from osu!. Could you add rendering snaking sliders as an option?

itdelatrisu:
Yeah, I've been meaning to do this but still haven't gotten around to it... if you have time, it'd be really helpful if you (or anyone) could look into the mechanics of it too -- e.g. how long snaking takes relative to (probably) approach time, or if there'd be issues with the current fade-in implementation (for instance, I had to rewrite that to get "Hidden" to work at all), etc.  I usually measure these kind of things by recording several videos and going through them frame-by-frame, which is really tedious but generally works.

I don't think I'll be able to implement this for the new slider style (mmsliders) though, but someone might be able to help out after the functionality is there (for peppysliders, at least).

Bigpet:
Yeah, I can do it for the "mmsliders" no problem. Once the timing is cleared up.

As for the mechanics. I assume that they snake in at the same rate as the "ball" moves across it (e.g they take "sliderTime" to completely snake in).

itdelatrisu:
Awesome, thanks for offering! (And sorry I keep relying on you for this... XD)

I doubt the snaking time is related to the slider time; if you think about those swirling sliders with ~50 control points, the slider time is far too long.  I only did a quick test, but on an Easy map, I got ~506ms for snaking for a ~933ms slider.

PR #134, opened 2015-09-16
Snaking sliders initial implementation.

Bigpet:
regarding #133 

Not ready for merging yet (mmsliders needs to be optimized). This is mainly for discussion around things like the pop-in of "repeat" arrows and other complications.

Draws curves in a range from 0 to x (where x is a value between 0 and 1), instead of always drawing the full curve

### DIFF
--- a/src/itdelatrisu/opsu/GameData.java
+++ b/src/itdelatrisu/opsu/GameData.java
@@ -907,7 +907,7 @@ public class GameData {
 						float oldColorAlpha = hitResult.color.a;
 						Colors.WHITE_FADE.a = alpha;
 						hitResult.color.a = alpha;
-						hitResult.curve.draw(hitResult.color);
+						hitResult.curve.draw(hitResult.color,1.0f);
 						Colors.WHITE_FADE.a = oldWhiteAlpha;
 						hitResult.color.a = oldColorAlpha;
 					}
--- a/src/itdelatrisu/opsu/Options.java
+++ b/src/itdelatrisu/opsu/Options.java
@@ -417,6 +417,7 @@ public class Options {
 		BACKGROUND_DIM ("Background Dim", "DimLevel", "Percentage to dim the background image during gameplay.", 50, 0, 100),
 		FORCE_DEFAULT_PLAYFIELD ("Force Default Playfield", "ForceDefaultPlayfield", "Override the song background with the default playfield background.", false),
 		IGNORE_BEATMAP_SKINS ("Ignore All Beatmap Skins", "IgnoreBeatmapSkins", "Never use skin element overrides provided by beatmaps.", false),
+		SNAKING_SLIDERS ("Snaking sliders", "SnakingSliders", "Sliders gradually snake out from their starting point.", true),
 		SHOW_HIT_LIGHTING ("Show Hit Lighting", "HitLighting", "Adds an effect behind hit explosions.", true),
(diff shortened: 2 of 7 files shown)


itdelatrisu:
Oh wow, that was fast, thanks!

The repeat arrows in osu! seem to fade in after the curve is fully faded in/expanded, and very quickly (around 100-200ms, but it's kind of hard to tell).  The ticks also seem to fade in at the same rate/time as the repeat arrows -- though maybe starting around 50ms later (again, it's hard to tell).  While you're at it, if you want, the ticks also get scaled from half size to full size while fading in, and with the `AnimationEquation.OUT_BACK` easing.  Everything else looks fine to me.

There should be an on/off flag in the game options for this, so just add this line:
```
SNAKING_SLIDERS ("Snaking sliders", "SnakingSliders", "Sliders gradually snake out from their starting point.\nThis should run fine unless you have a low-end PC.", true),
```
And I'd add it right below the `IGNORE_BEATMAP_SKINS` option in `OptionsMenu.java`.

Lastly, I've just been calling every [0,1] parameter `t`, and I'd clamp all the `drawUpTo` definitions to the interval be safe.  There's also a Javadoc warning on `CurveRenderState.java:84` in a `@link` tag since you changed a method signature.

Bigpet:
I'm not so sure about the performance impact with optimizations.

It might be a toss up or a trade-off. Since now the spike in draw calls when a new curve is rendered is lessened a bit (since the whole curve does not need to be drawn in 1 frame), but there is more frames in which the framebuffer needs to be switched to draw the next few segments of the slider.

snakingperformance

(btw. I still need to do some cleanup and the return arrow blend in, so not ready for merging)

itdelatrisu:
Do you think performance will actually be an issue?  And I don't think it's much of a problem anyway, since we'll have the on/off switch.

Let me know if there's any part of this you want me to do, too.

Bigpet:
@itdelatrisu no, I don't think performance will be an issue, hence my comment. I was just trying to say that "This should run fine unless you have a low-end PC" might be inaccurate and there might actually be some more stuttering without snaking slider in some cases.

itdelatrisu:
Oh, sorry, I misread that then.  We can take that line out; I just copied the text from osu!.

Bigpet:
Alright, I think I got the most important things down.

You can start reviewing I guess.

Just one question, do you want `0.5f + 0.5f*AnimationEquation.OUT_BACK.calc(decorationsAlpha);` or `AnimationEquation.OUT_BACK.calc(0.5f+0.5f*decorationsAlpha);` ?

Also, I might have asked this once already, but do you prefer things squashed into one commit?

itdelatrisu:
I think it looks good.  I'd just rewrite the tick code to avoid recalculating values:
```
		if (ticksT != null) {
			float tickScale = 0.5f + 0.5f * AnimationEquation.OUT_BACK.calc(decorationsAlpha);
			Image tick = GameImage.SLIDER_TICK.getImage().getScaledCopy(tickScale);
			Colors.WHITE_FADE.a = decorationsAlpha;
			for (int i = 0; i < ticksT.length; i++) {
				Vec2f c = curve.pointAt(ticksT[i]);
				tick.drawCentered(c.x, c.y, Colors.WHITE_FADE);
			}
			Colors.WHITE_FADE.a = alpha;
		}
```
I don't really care if you squash or not.  I'll merge this shortly if you're done.  Thanks for doing this!

Bigpet:
@itdelatrisu yeah, I think that should be it for now.

PR #148, opened 2015-11-16
Make music pitch and speed go down on fail (like in osu!)

mad-s:


itdelatrisu:
Thanks!  This works, but it causes other problems -- the music doesn't continue playing after you fail.  (If you hit retry, nothing happens unless you hit "skip".  If you go back to the song menu, the music doesn't play until you select a new song.)

mad-s:
Guess I haven't tested enough. I'll work on it.

PR #150, opened 2015-11-18
Add falling game objects on fail

mad-s:


itdelatrisu:
This looks great -- thanks a lot!

Tell me if these were intentional, but here are the visual differences I noticed from osu!:
* There's no fade-out to the fail menu in osu!.
* osu! still shows hit misses when objects start falling (though hit misses are in the original locations).

I'm not able to test the audio part of this at the moment, but will probably get a chance in a few hours.

mad-s:
That would not be too difficult to change. I can't really compare to osu! since I haven't gotten it to work in Wine, but I find the current version quite appealing.

itdelatrisu:
Yeah, I do think it looks pretty good as it is.  I think I'd prefer no fade-out or a much shorter one, though, but I don't really care about the hit misses (you can decide for yourself).

mad-s:
The transition is just for convenience since it blocks input. You could define your own transition with a delayed fade.

PR #152, opened 2015-11-22
Draw reversing arrow in black if the slider is lightly colored

mad-s:
On some maps the reversing arrows can be really difficult to see on white sliders.


itdelatrisu:
Looks good.  Thanks again!

PR #155, opened 2015-11-29
Add ingame scoreboard

mad-s:


itdelatrisu:
Thanks a lot!  I really appreciate all the help -- I've been meaning to add a bunch of the features you've been writing for a long time.

Would you mind improving the appearance of this before I merge it, though?  (I can do this if you don't want to, though I don't think I'll have any spare time for a while -- just let me know.)  For reference, here's how osu! looks now:
image
So, a semi-transparent background (image?), some padding, commas in the numbers, and changing the combo text placement would do it.

Also, as for fading in/out: in osu!, the scoreboard fades smoothly when pressing tab, and it will always be displayed during breaks regardless (though I don't really care for this last part).

Lastly, this happens when there are a lot of clustered scores: https://www.youtube.com/watch?v=9TdOLlAhg6w

mad-s:
Yes, I can improve it tomorrow afternoon. I think the clustered scores are too much of a problem, also it would require a more sophisticated animation system. Also one could decrease the animation time to mask the issue.

mad-s:
I made the scoreboard fade as well as being visible during breaks. A also added commas and moved the combo text to the right. The scoreboard can handle up to 999'999'999 score and 9'999x combo without overlapping or connecting, I think that is enough.

honard:
(I know this is a stupid question but)
When will you release the New Updated Android Version of opsu! ?

thanks for replying ^-^

iceblade112:
@honard Late reply, but @itdelatrisu doesn't actually maintain the Android version like the normal PC version.

itdelatrisu:
Sorry for the super late response on this -- finally had a chance to look at this, will merge and make a follow-up.  Thanks so much!

Issue #156, opened 2015-11-30
follow points improperly positioned when image is larger


azhao12345:
![opsubig](https://cloud.githubusercontent.com/assets/6671151/11462311/35144676-96c7-11e5-9dcc-225cbddd8201.PNG)

PR #157, opened 2015-11-30
centered follow points

azhao12345:
#156


itdelatrisu:
Good catch.  Thank you!

PR #178, opened 2016-01-26
Attempt to fix issue 177

Nightflyyer:


mad-s:
I think `MusicController.setPitch(1.0);` would be more appropriate since it keeps the current position.

itdelatrisu:
Fixed by #186, thank you though!

PR #180, opened 2016-02-17
More consistent Music Offset for most beatmaps.

fluddokt:
(My accuracy sucks so kinda hard to test)


mad-s:
The display could be quite intrusive during gameplay and can't even be disabled. Also I don't think we should show any numbers without any explanation as to what they mean.
Maybe just show the average offset at the ranking screen together with an explanation *if* the player has the hit error bar enabled. Definitely don't show the information during gameplay, we already have the hit error bar.

PR #205, opened 2016-12-03
Improve slider rendering.

mad-s:
The old code created a framebuffer for every slider drawn, which results in a major performance loss, especially on shared memory systems like integrated GPUs.
The new codes renders a 'ramp', not individual cones, so the sliders look extra smooth.
Also, there is no framebuffer involved, meaning the new sliders can be used on more systems.

### DIFF
--- a/src/itdelatrisu/opsu/GameData.java
+++ b/src/itdelatrisu/opsu/GameData.java
@@ -389,7 +389,7 @@ public class GameData {
 		if (hitResultList != null) {
 			for (HitObjectResult hitResult : hitResultList) {
 				if (hitResult.curve != null)
-					hitResult.curve.discardCache();
+					hitResult.curve.discardGeometry();
 			}
 		}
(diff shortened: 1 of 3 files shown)


itdelatrisu:
Awesome, this looks way better.  I'll wait a few days before merging in case anyone else wants to comment.  Thanks a lot!

Lignumm:
With this change, the slider curve does not render for me anymore. It just looks like this:


I'm on Arch (Linux 4.8.11) with a GTX 760Ti. My OpenGL version string is `4.5.0 NVIDIA 375.20`.

On a side note, does [this glFlush call](https://github.com/itdelatrisu/opsu/blob/999136407701513b65a20d7d4782b79538c16172/src/itdelatrisu/opsu/render/CurveRenderState.java#L496) serve any purpose? It looks redundant to me.

mad-s:
Sorry, I'm not used to writing GLSL 1.1. Should be fixed now.

PR #207, opened 2016-12-08
Added GL version to error handler

Lemmmy:


itdelatrisu:
If there's no OpenGL context when this fires, you'll get a `RuntimeException` on that line:
```
java.lang.RuntimeException: No OpenGL context found in the current thread.
	at org.lwjgl.opengl.GLContext.getCapabilities(GLContext.java:124)
	at org.lwjgl.opengl.GL11.glGetString(GL11.java:1792)
```

Lemmmy:
Does that fix it?

itdelatrisu:
Have you gotten this to work?  I don't think we can get the value in that class, since a lot of exceptions (esp. uncaught exceptions caught in `Thread.setDefaultUncaughtExceptionHandler()`) seem to run in a different thread (i.e. without the OpenGL context).

itdelatrisu:
Done in af667a4. Thanks!

Issue #167, opened 2015-12-21
[Feature Request] Reverse Arrow,And Slider followcircle


honard:
Can u add Bumping effects in the reverse arrows in a slider?

and about the Slider Followcircle 
can u add the featurewhen starting to follow the slider,the slider follow circle appears with soft kinetic movement,and when the slider is finished ,the slider follow circle will shrink until it fits to the Slider end circle

thanks for the feedback

Lemmmy:
@itdelatrisu is this issue finished with #210?

itdelatrisu:
Only the bouncing reverse arrow was addressed in #210 but I took it out since it looks dumb with the current reversearrow image (or any that fit the entire hit circle).

Took care of all follow circle animations in 99c698b.

PR #210, opened 2016-12-10
Slider improvements (hidden, followcircle, ticks, reversearrow)

yugecin:
see also #167


itdelatrisu:
Awesome!  Thanks for noticing these things, this looks great.

I personally don't like the appearance of the bouncing reverse arrow, at least with the default skin; it's too distracting unless we switch to an image with a transparent background.  I'm going to comment it out for now.

PR #211, opened 2016-12-12
Slider improvements (fade in ticks, fade out reversearrow, animations)

yugecin:
Fade out reverse arrow when passed repeat
Fade in sliderticks when passed repeat

### DIFF
--- a/src/itdelatrisu/opsu/GameData.java
+++ b/src/itdelatrisu/opsu/GameData.java
@@ -141,16 +141,17 @@ public class GameData {
 
 	/** Hit result types. */
 	public static final int
-		HIT_MISS     = 0,
-		HIT_50       = 1,
-		HIT_100      = 2,
-		HIT_300      = 3,
-		HIT_100K     = 4,   // 100-Katu
-		HIT_300K     = 5,   // 300-Katu
-		HIT_300G     = 6,   // Geki
-		HIT_SLIDER10 = 7,
-		HIT_SLIDER30 = 8,
-		HIT_MAX      = 9;   // not a hit result
+		HIT_MISS          = 0,
+		HIT_50            = 1,
+		HIT_100           = 2,
+		HIT_300           = 3,
+		HIT_100K          = 4,   // 100-Katu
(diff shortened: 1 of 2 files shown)


itdelatrisu:
Thanks again!  I'm a little hesitant on fading out the reverse arrows without implementing the full related behavior -- the hitcircle/hitcircleoverlay sprites also fade out when that point on the slider is done with.  Otherwise, this is kind of distracting on its own.

yugecin:
Yep, will do

itdelatrisu:
Yeah, that's right (I think).  From looking at osu! a few more times, this is what I can gather:
- The circles start fading out right after the time has passed (if it won't get hit again with repeats), even if you *weren't* holding the slider.
- If you *were* holding the slider, there should be a "pop" effect (fade+scale) for the hit circles.  This should be at every point on the slider (e.g. first point, any repeats, last point).  Note: currently, opsu! has a "pop" only at the end of a slider, and it's actually implemented wrong -- it currently pops if the **slider score** was not a miss, but it should really only consider if you were holding the slider **at that point**.
Thanks so much for helping out!

Off-topic: I just noticed your cursordance project, it's really cool! :)

yugecin:
I recorded some gameplay and watched it frame by frame, these are my notes:
* holding slider
  * initial circle: pop
  * end circle: pop
  * return point: pop reverse arrow & pop circle
* not holding slider
  * initial circle: fade out
  * end circle: disappear
  * return point: reverse arrow disappears & circle disappears
* when a circle popped/faded/disappeared, fade in a new circle (and reversearrow if needed) if the ball has to return to this point again, seems like alpha is 1f at the first slidertick (should check with sliders without sliderticks)

--
Thanks :D I'll add it to opsu! too when I have the chance.

yugecin:
Should be ready to go :)

itdelatrisu:
This looks amazing! Thanks so much! 😄

About the sliders cutting off - we can wait for a reply (in #215), or you could make a temporary fix to always draw the end circle and we can revert it later.  What would you prefer?

PR #212, opened 2016-12-14
Fix falling sliders with new slider style.

mad-s:
Also changed it so that game objects rotate around their center and not their starting point.


itdelatrisu:
Yay, thanks a lot!

PR #216, opened 2016-12-19
Fix #215.

### LINKED ISSUE #215: Snaking sliders are cut off
Snaking sliders should be rounded, but are cut off (in #205).  This is noticeable with the higher-level slider changes in #211, when the end hit circle is faded in after snaking completes (as in osu!).

@mad-s, how difficult would it be to change this?

mad-s:
I needed to slightly bump GLSL version requirements (for gl_VertexID), from GLSL 1.10 to 1.30, in turn bumping OpenGL requirement from 2.0 to 3.0.


itdelatrisu:
Looks good to me.  Thanks for the quick fix!

Issue #231, opened 2017-01-07
Add option to disable ranking elements animation


tpenguinltg:
Commit e49ca3c introduced animation for the game ranking screen elements. I for one prefer to see my results instantly instead of having to wait for the animation or clicking to finish the animation every time because it allows me to move on to the next song sooner.

itdelatrisu:
Eh... hesitant on adding an option that saves one mouse click. :/  I don't see anyone else requesting this on osu! either.

tpenguinltg:
I would be okay with shortening the animation time (perhaps cut it down to 3 seconds instead of 6). I hear it isn't 6 seconds in osu!, but I have not been able to verify this myself as I don't have a copy.

itdelatrisu:
It is 6 seconds, I timed it pretty closely to osu!: https://youtu.be/dvmJqij6LbE

tpenguinltg:
The animation in osu! feels faster, even if they're both 6 seconds. It's probably because the score display increases while the animation happens.

itdelatrisu:
You're free to make a PR with the score changing animation.  I'm not exactly sure how it works in osu!, but it looks like it picks random digits and gradually sets the correct digits from right to left.

Will close this issue for now, since I don't really feel this option would be useful to many other people.  Thanks though!

PR #233, opened 2017-01-07
Animate score display in the ranking screen

tpenguinltg:
Animate the score display in the ranking screen, finishing when the white flash appears. It displays the correct digits from left to right in equal time intervals, filling the rest of the digits with random digits.

This is for parity with osu!.
See also #231.


itdelatrisu:
Looks great! Thanks!  Surprised Java doesn't crash on log(0). :)

Issue #89, opened 2015-06-07
Failed to load file 'combobreak.wav'.


rodrigezBCL:
**Version:** 0.9.0
**Build date:** 2015-05-17 19:41
**OS:** Linux (i386)
**JRE:** 1.7.0_79
**Error:** `Failed to load file 'combobreak.wav'.`
**Stack trace:**

```
java.lang.IllegalArgumentException: Invalid format
    at org.classpath.icedtea.pulseaudio.PulseAudioDataLine.createStream(PulseAudioDataLine.java:142)
    at org.classpath.icedtea.pulseaudio.PulseAudioDataLine.open(PulseAudioDataLine.java:99)
    at org.classpath.icedtea.pulseaudio.PulseAudioDataLine.open(PulseAudioDataLine.java:283)
    at org.classpath.icedtea.pulseaudio.PulseAudioClip.open(PulseAudioClip.java:402)
    at itdelatrisu.opsu.audio.MultiClip.getClip(MultiClip.java:164)
    at itdelatrisu.opsu.audio.MultiClip.<init>(MultiClip.java:94)
    at itdelatrisu.opsu.audio.SoundController.loadClip(SoundController.java:117)
    at itdelatrisu.opsu.audio.SoundController.loadClip(SoundController.java:90)
    at itdelatrisu.opsu.audio.SoundController.init(SoundController.java:207)
    at itdelatrisu.opsu.states.Splash$1.run(Splash.java:109)
```

itdelatrisu:
Can you upload or link to the files that are causing the errors?

rodrigezBCL:
I just put the setting "Disable all sound effects" to No and a get all .wav effects failed to load. Putting on Yes it don't happen maybe it's my Linux sound driver

Bigpet:
That sound file is a 8-bit 11025Hz mono wav file and it seems like icedtea just uses pulseaudio directly to do the audio output and the pulse documentation says that it should support 8-bit PCM and sample rates anywhere from 1Hz to 192000Hz.

I tried to run it with icedtea on my Ubuntu lappy and I actually get a different exception (line unavailable).

I do wonder though @itdelatrisu , should we maybe just convert all sound samples to 44100Hz 16-bit stereo PCM? Having them all be different seems kind of weird and the re-sampled sounds would only be around 2-3 times the size they currently are.

Bigpet:
[...]

PR #238, opened 2017-01-22
Fix Sound Effects on Linux

chanceVermilion:
First off I want to tell you that I just found this project and I'm impressed. I haven't played Osu! in years and in the mean time I've made Ubuntu my OS of choice so I was happy to find a native way to play the game again.

This fixes (on my Ubuntu 16.04 machine) #89 . Of course that fix introduced another issue on the Download Menu which I fixed by removing what looks like a redundant call to stop the track. I also did a directory restructure to make the project more maven-y, added a test, and changed the exec plugin to run during the integration-test phase. I've played around with this on one machine and now the sound effects and song download previews work.


itdelatrisu:
Thanks a lot for figuring this out!  (Especially since I never managed to reproduce any of the audio issues on Linux...)

I committed just the audio fix in 039dd66 (didn't want to bother you by asking to not move around imports, etc.).

I'm slightly against the other changes for these reasons:
* Not really a fan of Maven's project structure with all the useless (for this project, at least) nested directories.  I think it's plenty clear enough as is, too.
* At this point it seems kind of hopeless/not worth the effort to add unit tests, and I don't think it makes sense to have a testing framework when there will inevitably be almost nothing added to it.  (Your test also wouldn't have failed on my (Windows) machine anyway.)
* I realize this is a pretty sketchy use of `mvn compile`, but `exec` isn't really an "integration test" either.  Not sure if there's a better place to put it. :/

Closing this PR since the main issue was addressed -- feel free to keep commenting if you disagree with anything I said.

PR #244, opened 2017-02-01
Back button

yugecin:
I like the default back button in osu!, so I made it for opsu! :)

At `BackButton.java:83` I check if the back button is skinned by looking at the size of the image, as the commented line above it doesn't seem to work like I thought it would, not sure why.


itdelatrisu:
Looks great! :) Thanks again!

Could you pull the latest opsu! commits into this?

Comments:
* The reason the commented line doesn't work is because `MENU_BACK` isn't set to be preloaded, and nothing loads the image before you call `hasGameSkinImage()`.  You should just set the `preload` flag in `GameImage.java`: (note - I added this constructor a few commits down from where you branched)
```
MENU_BACK ("menu-back", "menu-back-%d", "png", false, true),
```
* The chevron in osu! pulses differently from the logo -- it's an in-out movement instead of exactly "pulsing" on the beat.  I think this would look closer, feel free to adjust it though:
```
float t = (beatProgress < 0.5f) ?
	AnimationEquation.OUT_BACK.calc(beatProgress * 2f) :
	1f - AnimationEquation.LINEAR.calc((beatProgress - 0.5f) * 2f);
int chevronSize = (int) (chevronBaseSize - (isHovered ? 6f : 3f) * t);
```

yugecin:
Should be ready to go :)
Didn't notice the chevron, I changed it to IN_QUAD & OUT_BACK

itdelatrisu:
Tweaked it a bit more, still not accurate but I tried.

PR #262, opened 2017-02-27
Show restart button for options that require a restart

tpenguinltg:
Add a new method `isRestartRequired` to the `GameOption` enum that returns `true` if the option requires a restart to take effect. The method returns `false` by default and the option should override this if `true` is needed. This approach is less messy than adding a flag to the various constructors.

The restart button was previously not shown for the following options:
- `DISABLE_SOUNDS`
- `ENABLE_WATCH_SERVICE`

The restart button continues to be shown for the following options:
- `SCREEN_RESOLUTION`
- `FULLSCREEN`
- `SKIN`


itdelatrisu:
Looks good, thanks!

Issue #266, opened 2017-03-11
Experimental slider tracks not falling on failure


tpenguinltg:
With experimental sliders and merging sliders enabled, the slider tracks don't fall on failure. They also continue to shrink if shrinking sliders are enabled (may or may not be correct, because balls still continue to roll with regular sliders).

With shrinking sliders:
![screenshot_2017-03-10_20-46-30](https://cloud.githubusercontent.com/assets/8697332/23819382/70766236-05d3-11e7-94e5-8b22008b6a89.png)

Without shrinking sliders:
![screenshot_2017-03-10_20-49-40](https://cloud.githubusercontent.com/assets/8697332/23819385/98a0506e-05d3-11e7-9ccc-0034f4081d90.png)

They fall as expected when merging sliders is disabled.

opsu! commit 8dd92de

itdelatrisu:
Thanks for reporting. Tagging @yugecin. (I might not have a chance to look at this for a while. x_x)

tpenguinltg:
I think I see the problem. It looks like the merged sliders are being drawn directly instead of being added to the hit object stack ([Game.java:1633](https://github.com/itdelatrisu/opsu/blob/06370bd055ac37f23562500bbf173191558f543a/src/itdelatrisu/opsu/states/Game.java#L1633)), so the transformations to make it fall aren't being done on it. I'm not quite sure how to fix this, though, because a `Graphics` object isn't being passed to the merged slider's `draw` method. I assume that the merged slider should fall as a unit rather than breaking up into its component parts.

yugecin:
The solution is to force non-merged sliders when they are falling, making fix atm.

PR #268, opened 2017-03-19
Issue 266

yugecin:
fix for #266, by forcing non-merged sliders when the game is in losing state.
* Also added a check to not draw sliderticks when slider is finished, it caused the ticks being shown when the slider should be finished when the objects are falling when losing.
* When losing with shrinking sliders enabled, the initial circle keeps showing while the slidertrack may already have disappeared. The shrinking seems okay as the sliderball also continues to moves (as pointed out by @tpenguinltg), but it might look a bit weird because the initial circle stays, not sure if it needs to be changed.

### DIFF
--- a/src/itdelatrisu/opsu/objects/Slider.java
+++ b/src/itdelatrisu/opsu/objects/Slider.java
@@ -377,6 +377,11 @@ public class Slider implements GameObject {
 	 * @param decorationsAlpha the decorations alpha level
 	 */
 	private void drawSliderTicks(int trackPosition, float curveAlpha, float decorationsAlpha) {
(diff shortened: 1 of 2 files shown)


tpenguinltg:
I had some doubts on whether this "easy way out" that I described would feel right, but it feels natural enough. I'd say this works.

tpenguinltg:
I think the slider start circle staying is fine also because if you miss the head of a slider during gameplay, the start circle stays until it's considered too late (the next beat?).

itdelatrisu:
This looks good to me.  Thanks a lot for fixing this! :)

Issue #277, opened 2017-04-02
Possible solutions for bloodcat


fluddokt:
I guess an easy solution would be to just open the page in the browser.
But 
-the user would have to choose to download to the Import directory so opsu could import it
-loses the internal download progress.
 Though it could continuously check the current filesize if they saved it the import directory for progress but still need the find the end filesize somehow.
-still provides the benefit of avoiding duplicate beatmaps at least.

Another way would be to somehow integrate the reCAPTCHA.
Not really how sure how possible this is. (Can at least get the page in the error stream) 

Might be able to do this with osu website also. (Open in browser  or  integrate login somehow)
https:// osu.ppy.sh/d/%d[n]

tpenguinltg:
I don't think it would be very easy to integrate reCAPTCHA without having to import a Web renderer (including a JS engine), and I think that would be outside the scope of this project.

I just tested another possible solution: passing in a session cookie. When you solve the captcha, it stores a session cookie so that you don't have to solve it again. It seems it doesn't validate the originating IP address, so you can successfully download from any IP address. The only thing I'm not sure about is how long the session lasts for. The cookie expires after 3 days, but that doesn't necessarily mean that the session does, especially if requests are made before the 3 days ends. More testing will need to be done.

fluddokt:
Seems interesting but I'm not sure what to do with that cookie, since it probably expires eventually anyways unless the user can somehow get their own cookie into opsu themselves.

tpenguinltg:
Actually, a cookie is only a "key=value" pair, and expiration is only a thing in a browser. You can send the same "key=value" pair with your request long after the browser cookie expires and it will still be a valid cookie. Whether bloodcat will catch on is uncertain, but it is a valid concern.

tpenguinltg:
As for osu! login, it's probably much simpler. It looks like it's just a POST request to `https://osu.ppy.sh/forum/ucp.php?mode=login` with fields `username` and `password` (and possibly also `redirect=/`, `sid`, and `login=login`), then you'd just keep the cookie around while the program is running. The username and password would be stored in the config file (I think osu! keeps it in plaintext anyway).

fluddokt:
Huh I always thought the server would expire the session also. Still seems like a pain if it does expire.
Yea osu login is totally possible cool.

itdelatrisu:
The cookie approach seems interesting (didn't know reCAPTCHA worked like that), but problems I have with approach:
[...]

PR #278, opened 2017-04-03
Download "Open in browser" Test #277 

fluddokt:
#277 
Something like this?
Quite hacky and not really done yet.
I'm thinking of adding a button to select the "import"/"another import" directory based on tpenguinltg idea.

### DIFF
--- a/src/itdelatrisu/opsu/Utils.java
+++ b/src/itdelatrisu/opsu/Utils.java
@@ -421,14 +421,16 @@ public class Utils {
 	 * @param file the ZIP archive
 	 * @param dest the destination directory
 	 */
-	public static void unzip(File file, File dest) {
+	public static boolean unzip(File file, File dest) {
 		try {
 			ZipFile zipFile = new ZipFile(file);
 			zipFile.extractAll(dest.getAbsolutePath());
 		} catch (ZipException e) {
-			ErrorHandler.error(String.format("Failed to unzip file %s to dest %s.",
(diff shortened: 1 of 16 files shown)


src/itdelatrisu/opsu/Utils.java
```diff
@@ -421,14 +421,16 @@ else if (doReplace)
 	 * @param file the ZIP archive
 	 * @param dest the destination directory
 	 */
-	public static void unzip(File file, File dest) {
+	public static boolean unzip(File file, File dest) {
```
Lemmmy: make sure to javadoc the `@return`

src/itdelatrisu/opsu/downloads/DownloadNode.java
```diff
@@ -263,29 +265,49 @@ public DownloadNode(int beatmapSetID, String date, String title,
 	 * @param server the server to download from
```
Lemmmy: update the javadoc to say `dlserver`

itdelatrisu:
Sorry for the terribly late response. Did something simpler in ab39b3e; I don't think there's much to gain from creating fake download status nodes when we're not really monitoring the download status.

As for separating download/search servers, I still feel like it'd be good for improving reliability, but I don't want users to get confused by too many buttons.  I don't think the tradeoff is worth it at the moment, but will reconsider it if the situation changes (ex. we're left with only one download server with no search API).

PR #281, opened 2017-04-10
Add options button to main menu

### LINKED ISSUE #279: Options button should be shown on main menu
The button to show the options overlay should be shown in the main menu. If the user starts opsu! with an empty beatmap database (e.g. when starting for the very first time), there is no way to get to the options without downloading a beatmap first to get to the songs menu. One case where this matters is if a Linux user is having problems with the sound effects and needs to disable it (I ran into this issue when testing something else earlier). I'm sure there are other use cases for it, and if I'm not mistaken, osu! has the options overlay accessible from the main menu.

tpenguinltg:
Adding this button to the main menu allows a new user (i.e. one who has no beatmaps) to access the options without installing a beatmap.
We may also want to consider making a new image so that it doesn't say "Other Options" when there are no other options on that screen.

Fixes #279

### DIFF
--- a/src/itdelatrisu/opsu/states/MainMenu.java
+++ b/src/itdelatrisu/opsu/states/MainMenu.java
@@ -925,6 +925,12 @@ public class MainMenu extends BasicGameState {
 		case Input.KEY_DOWN:
 			UI.changeVolume(-1);
 			break;
+		case Input.KEY_O:
+			SoundController.playSound(SoundEffect.MENUHIT);
+			showOptionsOverlay = true;
+			optionsOverlayProgress.setTime(0);
+			optionsOverlay.activate();
+			break;
 		}
 	}
 
--- a/src/itdelatrisu/opsu/states/MainMenu.java
+++ b/src/itdelatrisu/opsu/states/MainMenu.java
@@ -29,7 +29,9 @@ import itdelatrisu.opsu.beatmap.Beatmap;
 import itdelatrisu.opsu.beatmap.BeatmapSetList;
 import itdelatrisu.opsu.beatmap.BeatmapSetNode;
 import itdelatrisu.opsu.downloads.Updater;
+import itdelatrisu.opsu.options.OptionGroup;
 import itdelatrisu.opsu.options.Options;
+import itdelatrisu.opsu.options.OptionsOverlay;
 import itdelatrisu.opsu.states.ButtonMenu.MenuState;
 import itdelatrisu.opsu.ui.Colors;
 import itdelatrisu.opsu.ui.Fonts;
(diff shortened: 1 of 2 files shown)


src/itdelatrisu/opsu/states/MainMenu.java
```diff
@@ -720,6 +783,15 @@ public void mousePressed(int button, int x, int y) {
 			return;
 		}
 
+		// options button actions
+		if (selectOptionsButton.contains(x, y)) {
```
yugecin: put this in `mouseClicked` to solve the disappearing problem.
You'll also need to add 
```
if (showOptionsOverlay || !optionsOverlayProgress.isFinished()) return;
```
at the top of `mouseClicked`

yugecin:
The CTRL+O shortcut should be added aswell (like in SongMenu)

tpenguinltg:
@yugecin: Thanks for reviewing. Would it matter if it's just O instead of Ctrl+O? I think the only reason it's Ctrl+O in SongMenu is because O would conflict with search.

tpenguinltg:
I just noticed that searching is broken. The search field seems to lose focus every time a key is pressed, and it doesn't even get focus when using the keyboard shortcut to open the overlay. Any ideas?

src/itdelatrisu/opsu/states/MainMenu.java
tpenguinltg: Looks like that fixed it. Thanks.

yugecin:
Something weird is happening with the events. Seems like the search textfield consumes the events in these cases, but not when the options menu is opened via the song menu.
No idea why it does that, but I think I found a workaround/fix that seems to be working by playing with the focus of the field:

Remove line https://github.com/itdelatrisu/opsu/blob/8dd92de527213c910141d43c0be5ac152247df5e/src/itdelatrisu/opsu/options/OptionsOverlay.java#L368
and add `searchField.setFocus(false);` after the key press was sent: https://github.com/itdelatrisu/opsu/blob/8dd92de527213c910141d43c0be5ac152247df5e/src/itdelatrisu/opsu/options/OptionsOverlay.java#L1070

I propose CTRL+O because it's also how it's done in osu!, and it makes it consistent with the keybind in the song menu (indeed to prevent conflict with search).
In osu! the O key is also used to open the main menu (like when you click the logo), so that would cause conflicts aswell if that would be added to opsu!.

tpenguinltg:
Weird, but it seems to work. Nice job finding that.

I implemented that fix and changed the keybinding to Ctrl+O. I don't see any other issues with this, so unless someone else spots something, I think this is good to merge.

itdelatrisu:
Thank you!  Sorry for the late response.  I've committed my own patch (f4fa217) largely based on this, with these changes:
* Didn't like having the image there, so this is keyboard-only for now.
* Used the full osu! behavior for `O` vs. `CTRL+O` as @yugecin described above.
* Moved a bunch of things around.

PR #287, opened 2017-04-22
Fix regular snaking sliders

### LINKED ISSUE #286: HEAD currently has no snaking sliders
I've built from head in anticipation of testing #269 before it is released, and noticed sliders do not snake despite the setting. I have bisected it to two commits:
```
$ git bisect skip
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
0ad829189ca7972dbcea61381d0573b5088b68aa
3db9be9bbbfada52f380e0f9af278f3bf572a97a
We cannot bisect more!
```

tpenguinltg:
Fix snaking sliders not snaking with regular sliders (introduced by 0ad8291). This was a logic error.

isCurveCompletelyDrawn <==> (snaking => alpha == 1)
"The curve is completely drawn if and only if snaking implies alpha = 1".
A implies B is the same as (NOT A) OR B.

Fixes #286

### DIFF
--- a/src/itdelatrisu/opsu/objects/Slider.java
+++ b/src/itdelatrisu/opsu/objects/Slider.java
@@ -217,7 +217,7 @@ public class Slider implements GameObject {
 
 		boolean isCurveCompletelyDrawn;
 		if (!Options.isExperimentalSliderStyle()) {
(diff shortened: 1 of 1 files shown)


itdelatrisu:
Thanks for the fix, sorry I've been gone for so long.

Issue #130, opened 2015-09-12
Multitouch support


LemonLake:
The other day chong and I talked a little bit about adding multitouch support. We should add support for multiple cursors and whatnot so that opsu!android can work like osu!droid, i.e. one aiming finger, the others tap.

itdelatrisu:
Is this even possible in LWJGL?

fluddokt:
opsu!android simulates extra touches as right click at the first finger position, so I think it should be able to one aim finger and others tap.

FenekAlfa:
Indeed, I can do streams on the android version by pressing the first note with my right index and then dragging it while alternating left index and left middle for taps.

itdelatrisu:
@fluddokt: How did you implement that?  If it's possible without libGDX features, it'd definitely be nice to have (since I play on a touchscreen too :P).

fluddokt:
Well libgdx api easily provides multitouch support, however it still doesn't work with the desktop version
since its backend is still LWJGL.
My implementation is here.
https://github.com/fluddokt/opsu/blob/master/src/fluddokt/opsu/fake/StateBasedGame.java#L168

Also I really can't find any easy way of getting multitouch onto desktop.

itdelatrisu:
Yeah, that's what I thought; thanks for pointing out the code for it.  I'll close this then since it's not really feasible with the current framework.

LemonLake:
aight neat

PR #288, opened 2017-04-25
Multi Touch 

Asumine:
Would be great if the game had a multi touch setting like Osu!Droid so that streaming on Opsu can actually be down with just a touch screen


tpenguinltg:
Linking #130

Did you mean to create an issue instead? I highly doubt merging fluddokt/opsu is the right solution; they're separate for a reason.

Lemmmy:
what the fuck?????

itdelatrisu:
As mentioned, unfortunately multitouch isn't possible in the (dead) desktop framework, but the Android version does support multitouch.

PR #289, opened 2017-05-06
fix slidercaps being drawn when they shouldn't

yugecin:
Slidercaps are being draw on the initial circle when experimental sliders are enabled while the slider caps option is disabled (blame 0ad8291 from pull request #263).
Incorrect behavior can be easily seen on the osu! tutorial map on every slider with reverse arrows. 
The commit inside this pull request fixes that.


itdelatrisu:
Thanks!  Looks good. :)

Issue #283, opened 2017-04-10
Options overlay navigation


tpenguinltg:
This is probably going to be a bit of work, but I'll add this here for anyone who wants to take it on.

osu! has links on the left side of its options overlay that allow you to jump to a specific section. The highlight should reflect the current section (top-most visible?) even while scrolling. All links should be visible even at 800x600, but if it doesn't fit, it may be best to not show it at that resolution. I don't have a copy of osu! to verify, but I suspect that when searching, the links for the sections that don't return any results are either hidden or disabled.

[![options_base](https://cloud.githubusercontent.com/assets/8697332/24874691/90870872-1df3-11e7-986c-ecfc31b88e7e.jpg)](https://osu.ppy.sh/wiki/Options)

PR #298, opened 2017-05-28
Optionmenu navigation

yugecin:
Sidebar navigation for the option menu, as described in #283 

Some differences between this and osu!:
* The pink indicator fades in osu! when active group changes.
* This nav menu is scrollable in osu!. I tested on 800x600 and all items fit on screen so I didn't make it scrollable.

Notes:
* I made a mess in `MainMenu` & `SongMenu` for animations alpha. The option menu fades quicker than the navigation elements in the menu, so I had to adjust some things. It's now a bit messier as the fading animation is not handled in the `OptionsOverlay` class, but on other places.
* ~~I made some (shit) icons based on the ones in osu!, if anyone is experienced in making such images, _please_ make better ones ^^~~ Used font-awesome
* Section headers will be slightly transparant when they're not the active section on screen, like in osu!

Extra: Search filter won't accept new characters when there would be no results + added the little animation that osu! does when that happens.

_I didn't forget to commit the added resources for once, yey!_

~~**EDIT:** hold on, someone suggested to use font-awesome for the icons, I'll probably delete the commit with my shitty icons~~ done


itdelatrisu:
This looks amazing, thanks so much!  Only a few comments:
* Options overlay should be a little wider (about 40% of the screen in osu!, 35% in opsu!).
* When options in multiple categories all fit on one page (ex. with a search filter), clicking anything but the top category causes a slight jitter; it should do nothing at all.

tpenguinltg:
Nice work!

Like @itdelatrisu mentioned, I would suggest widening the overlay because the text in long lines get cut off (e.g. the option under Language) and short sliders like the "Track checkpoint" slider is also almost unusable.
screenshot_20170528_185010
screenshot_20170528_185615

yugecin:
Done & done!

The width was a silly mistake; it was 0.42x the width but the width of the nav bar was subtracted from it... I fixed that and made it 0.4x width for widescreen, and 0.5x for 4:3 resolutions (as in osu!).

itdelatrisu:
Thanks again! :)

Issue #276, opened 2017-03-30
Fail sound doesnt stop after exiting fail menu.


Conor29134:
Video:
https://www.youtube.com/watch?v=-fA6ZIxaxrA

itdelatrisu:
Noted, thanks for the report and video.  (I probably won't have time to get to this for a while though.)

Self-note: linking yugecin/opsu-dance#31, but I don't like their solution (relies on too many globals).

itdelatrisu:
Fixed in #299.

PR #299, opened 2017-05-30
Possible fix for failsound/applause not stopping

Lyonlancer5:
# Overview

Hi again, I've put up a possible fix for #276 (translations would have to wait until I get the new options overlay merged)

### Notes

The way it works rely on how `getClip()` works on the MultiClip class. On my setup, this solves the problem. 
It doesn't mute the clip but instead stops it so it can be played again, although not completely sure if this will work in other setups...

> Returns a Clip that is _not playing_ from the list.
> If no clip is available, then a new one is created _if under MAX_CLIPS_.
> Otherwise, an existing clip will be returned


itdelatrisu:
Works for me too.  Thanks!

PR #301, opened 2017-06-05
Use current score on in-game scoreboard

Lyonlancer5:
# Overview

The translation framework is _almost_ complete, so here's another thing I would suggest: how the sidebar scoreboard's score is updated. Instead of using the 'sliding' score for the in-game scoreboard, use the current score. This doesn't affect the sliding effect on the score display at the top-right corner of the game screen.

Although, the benefits apply only to score match multi-playing or other events that involve real-time/online scoreboard updating, both of which are not yet implemented.

Feel free to comment your opinions and make changes if you like.


tpenguinltg:
While I like the idea of using the current score, the transition doesn't feel fluid enough. I can't think of a good solution other than sliding.

If I'm not mistaken, the order of the in-game scoreboard is based on `currentScore`, which is currently the display score? Perhaps you can display the sliding score, but use the real score for the calculations, and hopefully the overhead of grabbing the score twice isn't too much. Of course, this might lead to split-seconds where the displayed score is less than the one below it, but it probably won't be noticeable unless you take a screenshot.

By the way, you may want to consider looking up how to do rebasing and squashing in Git (of particular interest is `git pull --rebase`). You have three commits for a single line change where you should really only have one.

itdelatrisu:
Worth mentioning that osu! uses the actual score, not sliding score.  I personally don't like it, and don't think it's that useful... so I'm going to keep it as is, and just add a comment that osu! does this differently.

PR #307, opened 2017-06-28
Key overlay

yugecin:
Notes
* osu! adds a text shadow, I couldn't find a decent way to do this (I commented a method to do this). This shadow is very useful when the text is colored white.
* I'm not sure how to correctly position/scale the background, it does not show correctly with some skins.


itdelatrisu:
Thanks as usual, looks great! But you forgot to commit the images. 😛

I'd prefer the `InputOverlayKey` class to go in a separate file (perhaps in `itdelatrisu.opsu.ui`) since `Game.java` is already way too long.

yugecin:
Whoops, added them (yugecin's law again >.<)
I also thought it should be in a separate file, but didn't know where to put it. Moved it now.

itdelatrisu:
Awesome, thanks! :)

PR #324, opened 2017-10-15
Add option "Unlimited" on frame limiter

Lyonlancer5:
Intended for use on benchmarking specific parts of the game, especially during render-intensive scenarios such as during high note densities, testing experimental sliders and the like, but can also be used to decrease input latency by indirectly affecting input polling rate.

Changes to input latency and FPS rates may vary depending on your system configuration.

### DIFF
--- a/src/itdelatrisu/opsu/options/Options.java
+++ b/src/itdelatrisu/opsu/options/Options.java
@@ -987,7 +987,7 @@ public class Options {
 	 * @param container the game container
 	 */
 	public static void setNextFPS(GameContainer container) {
-		int index = (targetFPSindex + 1) % (targetFPS.length - 1);
+		int index = (targetFPSindex + 1) % (targetFPS.length - 1); // Skip "Unlimited" option
 		GameOption.TARGET_FPS.selectItem(index, container);
 		UI.getNotificationManager().sendBarNotification(String.format("Frame limiter: %s", GameOption.TARGET_FPS.getValueString()));
 	}
--- a/src/itdelatrisu/opsu/options/Options.java
+++ b/src/itdelatrisu/opsu/options/Options.java
@@ -987,7 +987,7 @@ public class Options {
 	 * @param container the game container
 	 */
 	public static void setNextFPS(GameContainer container) {
-		int index = (targetFPSindex + 1) % targetFPS.length;
+		int index = (targetFPSindex + 1) % (targetFPS.length - 1);
 		GameOption.TARGET_FPS.selectItem(index, container);
 		UI.getNotificationManager().sendBarNotification(String.format("Frame limiter: %s", GameOption.TARGET_FPS.getValueString()));
 	}
--- a/src/itdelatrisu/opsu/options/Options.java
+++ b/src/itdelatrisu/opsu/options/Options.java
@@ -399,8 +399,7 @@ public class Options {
 
 			@Override
 			public String getValueString() {
-				return String.format((getTargetFPS() == -1) ? "Unlimited" : 
-					(getTargetFPS() == 60) ? "%dfps (vsync)" : "%dfps", getTargetFPS());
+				return String.format((getTargetFPS() == 60) ? "%dfps (vsync)" : "%dfps", getTargetFPS());
 			}
 
 			@Override
--- a/src/itdelatrisu/opsu/options/Options.java
+++ b/src/itdelatrisu/opsu/options/Options.java
@@ -399,7 +399,8 @@ public class Options {
 
 			@Override
 			public String getValueString() {
-				return String.format((getTargetFPS() == 60) ? "%dfps (vsync)" : "%dfps", getTargetFPS());
+				return String.format((getTargetFPS() == -1) ? "Unlimited" : 
+					(getTargetFPS() == 60) ? "%dfps (vsync)" : "%dfps", getTargetFPS());
 			}
 
 			@Override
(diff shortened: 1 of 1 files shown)


itdelatrisu:
I still don't really like this idea (was proposed a long time ago, #40).  Personally, I find it annoying when I'm toggling FPS and suddenly hit 100% CPU usage.  Towards your points, I don't believe that this game should be run past 240fps (which already seems high), and it's not the correct way to go for improving input latency.

LovelyA72:
Actually i can buy a pretty good 240hz monitor from taobao(an online shopping mall in China) for only 370 bucks. And if the max frame rate of opsu! is 240 fps, it only reach the standard frame rate of the monitor(v sync). And because osu! have this option for a long time and it is pretty stable. I recommend to add this feature.

Lemmmy:
I think it would be fine to have it as a setting, but not a keyboard shortcut?

Lyonlancer5:
I had proposed this because the game logic update and render update are both executing on one thread (AFAIK that's how updating the game's logic and display are implemented on Slick).

Any calls to `Thread.sleep()`, being blocked by I/O, long/heavy computations or anything that would be CPU-intensive would delay either of the updates.

For a long-term solution though, having to separate render and logic updates into different threads would be better.

yugecin:
> For a long-term solution though, having to separate render and logic updates into different threads would be better.
that's a pain to synchronize
you don't need separate threads, just don't render on every update https://github.com/yugecin/opsu-dance/blob/7a51828102ba94c72598d4fe66a4438f90f631aa/src/yugecin/opsudance/core/DisplayContainer.java#L246

itdelatrisu:
@Lemmmy: Okay, that's a good compromise. @Lyonlancer5, if you want to update this PR to do that, I'll accept it.

@LovelyA72: I understand, but do you really notice a difference with a >240hz refresh rate compared to 120 or 240 for a 2D game?

@yugecin: Do you find that doing that actually helps?  And with what kind of framerates / computer specs?  I'd be surprised if the Slick2D developers would overlook something as simple as this, which is why I haven't bothered trying it.

Lyonlancer5:
Noted of @Lemmmy's idea.
I do ask, what are your PC's current specs? It barely reaches 100% CPU usage on any of my setups (except for my Pentium 4 build, **that one** just maxes out even at ~120FPS)

src/itdelatrisu/opsu/options/Options.java
```diff
@@ -399,15 +399,17 @@ public void selectItem(int index, GameContainer container) {
 
 			@Override
 			public String getValueString() {
-				return String.format((getTargetFPS() == 60) ? "%dfps (vsync)" : "%dfps", getTargetFPS());
+				return String.format((getTargetFPS() == -1) ? "Unlimited" :
```
itdelatrisu: Take `(getTargetFPS() == -1) ? "Unlimited" :` out of `String.format()`.

src/itdelatrisu/opsu/options/Options.java
```diff
@@ -399,15 +399,17 @@ public void selectItem(int index, GameContainer container) {
...
+					(getTargetFPS() == 60) ? "%dfps (vsync)" : "%dfps", getTargetFPS());
 			}
 
 			@Override
 			public Object[] getItemList() {
 				if (itemList == null) {
 					itemList = new String[targetFPS.length];
 					for (int i = 0; i < targetFPS.length; i++)
-						itemList[i] = String.format((targetFPS[i] == 60) ? "%dfps (vsync)" : "%dfps", targetFPS[i]);
+						itemList[i] = String.format((targetFPS[i] == -1) ? "Unlimited" :
```
itdelatrisu: (same as above)

src/itdelatrisu/opsu/options/Options.java
```diff
@@ -985,7 +987,7 @@ private Options() {}
 	 * @param container the game container
 	 */
 	public static void setNextFPS(GameContainer container) {
-		int index = (targetFPSindex + 1) % targetFPS.length;
+		int index = (targetFPSindex + 1) % (targetFPS.length - 1);
```
itdelatrisu: Add a comment:
```
int index = (targetFPSindex + 1) % (targetFPS.length - 1);  // Skip "Unlimited" option
```

yugecin:
> @yugecin: Do you find that doing that actually helps? And with what kind of framerates / computer specs? I'd be surprised if the Slick2D developers would overlook something as simple as this, which is why I haven't bothered trying it.

It actually feels much more responsive when I do that, and I get a bit less 100s. It might also be a placebo effect (or something weird I did in my code), but for me it feels like it helps.

120 updates/s with 60fps feels much smoother than 60ups & 60fps

osu! (and other rythm games) really need fast input processing, whereas many other types of games may not, so that might be why they don't give options for that. It's probably not a common thing in games to do that actually (not that I know of).

```
          Processor: Intel(R) Core(TM) i7-2600 CPU @ 3.40GHz (8 CPUs), ~3.4GHz
          Card name: AMD Radeon (TM) R7 370 Series
```

src/itdelatrisu/opsu/options/Options.java
Lyonlancer5: Wait, what?

The shortcut function to "Unlimited" has already been removed. Removing this would result in the option "Unlimited" being ***not displayed on the option's dropdown list***, but available via modifying the configuration file's `FrameSync` parameter to -1.

src/itdelatrisu/opsu/options/Options.java
Lyonlancer5: Noted.

itdelatrisu:
Cleaned up and added in 6449b73.

Issue #302, opened 2017-06-06
[Aspire 2017] "Crashpire"


Leidi-Kana:
I found another Aspire map, _it didn't end well..._

This map: [Helblinde - The Solace of Oblivion (mapped by MinG3012)](https://osu.ppy.sh/s/594828)
<details>
<summary>Broken stuff</summary>
<ul>
<li>-1 star rating</li>
<li>Negative infinity BPM</li>
<li>~1.5k combo on some sliders for about a second</li>
</ul></details><br>

Upon **clicking it on song select**, the following exception is thrown:

<pre>
Sun Jun 4 01:24:49 CST 2017 ERROR:null
java.lang.NegativeArraySizeException
	at itdelatrisu.opsu.objects.curves.EqualDistanceMultiCurve.init(EqualDistanceMultiCurve.java:65)
	at itdelatrisu.opsu.objects.curves.LinearBezier.<init>(LinearBezier.java:83)
	at itdelatrisu.opsu.beatmap.HitObject.getSliderCurve(HitObject.java:430)
	at itdelatrisu.opsu.beatmap.tpSlider.<init>(BeatmapDifficultyCalculator.java:526)
	at itdelatrisu.opsu.beatmap.tpHitObject.<init>(BeatmapDifficultyCalculator.java:336)
	at itdelatrisu.opsu.beatmap.BeatmapDifficultyCalculator.calculate(BeatmapDifficultyCalculator.java:146)
	at itdelatrisu.opsu.states.SongMenu.calculateStarRatings(SongMenu.java:1972)
	at itdelatrisu.opsu.states.SongMenu.setFocus(SongMenu.java:1751)
	at itdelatrisu.opsu.states.SongMenu.moveFocus(SongMenu.java:1643)
	at itdelatrisu.opsu.states.SongMenu.keyPressed(SongMenu.java:1275)
	at org.newdawn.slick.state.StateBasedGame.keyPressed(StateBasedGame.java:445)
	at org.newdawn.slick.Input.poll(Input.java:1197)
	at org.newdawn.slick.GameContainer.updateAndRender(GameContainer.java:680)
	at itdelatrisu.opsu.Container.gameLoop(Container.java:136)
	at itdelatrisu.opsu.Container.start(Container.java:80)
	at itdelatrisu.opsu.Opsu.main(Opsu.java:213)
</pre>

<details>
<summary>On a low-spec PC, even vanilla osu! goes poof</summary>
[...]

PR #325, opened 2017-10-15
Aspire map patches (fixes #297, #302)

### LINKED ISSUE #297: [Aspire 2017]

Lyonlancer5:
## Overview
After having done roughly 3 months of research, diving into Lazer's C# code (thanks @sr229) and a few tests, playthroughs and watching Auto, I present a PR that would allow Aspire 2017 maps to load and be played without hard-crashing.

However, this is still incomplete:

#### Crashpire (#302)
- [x] Fix `NegativeArraySizeException` when evaluating certain sliders
- [x] Fix `NullPointerException` when a null curve is being loaded to a slider
- [x] Fix `ArrayIndexOutOfBoundsException` when attempting to call `pointAt(float)` where the value of the float is 1 but the `curve` array has a length of 1 (element 0)
- [x]  Parse timings correctly (currently skips nearly 2/3 of the song and note spamming after the 'Skip' portion)
- [x] Compute star rating as per new formula (osu-stable: 6.5 stars, osu-lazer: 130~ stars)

#### Slider issues (#297)
- [x] Fix slider freezing with experimental sliders enabled.
- [x] Fix non-disappearing sliders with experimental sliders disabled.
- [ ] Fix graphical glitching on "repeating" and "double-backing" sliders with experimental sliders disabled.

### DIFF
--- a/src/itdelatrisu/opsu/beatmap/BeatmapParser.java
+++ b/src/itdelatrisu/opsu/beatmap/BeatmapParser.java
@@ -36,6 +36,7 @@ import java.io.InputStreamReader;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -679,6 +680,7 @@ public class BeatmapParser {
 			return;
 
 		beatmap.objects = new HitObject[(beatmap.hitObjectCircle + beatmap.hitObjectSlider + beatmap.hitObjectSpinner)];
+		ArrayList<HitObject> objects = new ArrayList<>();
 
 		try (BufferedReader in = new BufferedReader(new FileReader(beatmap.getFile()))) {
 			String line = in.readLine();
@@ -732,12 +734,21 @@ public class BeatmapParser {
 					hitObject.setComboIndex(comboIndex);
 					hitObject.setComboNumber(comboNumber++);
 
-					beatmap.objects[objectIndex++] = hitObject;
+					objects.add(hitObject);
+					objectIndex++;
 				} catch (Exception e) {
 					Log.warn(String.format("Failed to read hit object '%s' for beatmap '%s'.",
 							line, beatmap.toString()), e);
 				}
 			}
+			objects.trimToSize();
+			Collections.sort(objects, new Comparator<HitObject>() {
+				@Override
+				public int compare(HitObject o1, HitObject o2) {
+					return o1.getTime() - o2.getTime();
+				}
+			});
+			beatmap.objects = objects.toArray(new HitObject[0]);
 
 			// check that all objects were parsed
 			if (objectIndex != beatmap.objects.length)
--- a/src/itdelatrisu/opsu/beatmap/BeatmapParser.java
+++ b/src/itdelatrisu/opsu/beatmap/BeatmapParser.java
@@ -35,6 +35,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -680,7 +681,6 @@ public class BeatmapParser {
 			return;
 
 		beatmap.objects = new HitObject[(beatmap.hitObjectCircle + beatmap.hitObjectSlider + beatmap.hitObjectSpinner)];
-		ArrayList<HitObject> objects = new ArrayList<>();
 
 		try (BufferedReader in = new BufferedReader(new FileReader(beatmap.getFile()))) {
 			String line = in.readLine();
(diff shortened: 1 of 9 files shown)


itdelatrisu:
Thanks for working on this! So these two maps load properly for me now, but trying to play them on "Auto" causes the game to freeze and use almost 100% of my CPU less than a minute into each map (Windows 10).  Is that not happening for you?

src/itdelatrisu/opsu/beatmap/BeatmapParser.java
```diff
@@ -732,12 +734,21 @@ public static void parseHitObjects(Beatmap beatmap) {
...
-					beatmap.objects[objectIndex++] = hitObject;
+					objects.add(hitObject);
+					objectIndex++;
 				} catch (Exception e) {
 					Log.warn(String.format("Failed to read hit object '%s' for beatmap '%s'.",
 							line, beatmap.toString()), e);
 				}
 			}
+			objects.trimToSize();
+			Collections.sort(objects, new Comparator<HitObject>() {
```
itdelatrisu: Is sorting really necessary? Does the osu! editor actually ever put the hit objects out of order (in time)?

Anyway, assuming there are incorrectly-ordered hit objects, sorting it like this would corrupt the combo index/number for the whole map.

src/itdelatrisu/opsu/beatmap/BeatmapParser.java
```diff
@@ -732,12 +734,21 @@ public static void parseHitObjects(Beatmap beatmap) {
...
 
-					beatmap.objects[objectIndex++] = hitObject;
+					objects.add(hitObject);
+					objectIndex++;
 				} catch (Exception e) {
 					Log.warn(String.format("Failed to read hit object '%s' for beatmap '%s'.",
 							line, beatmap.toString()), e);
 				}
 			}
+			objects.trimToSize();
```
itdelatrisu: There's no point calling this method for local variables. (It might even result in more space getting allocated before GC.)

src/itdelatrisu/opsu/beatmap/TimingPoint.java
```diff
@@ -83,7 +83,7 @@ public TimingPoint(String line) {
 
 		// tokens[1] is either beatLength (positive) or velocity (negative)
 		float beatLength = Float.parseFloat(tokens[1]);
-		if (beatLength > 0)
+		if (beatLength >= 0)
```
itdelatrisu: Explain this change? How can beatLength be 0?

sr229: there's probably instances a beatlength could be zero because again, during Aspire maps, mappers exploit the mapping system of osu! to do something really creative.

src/itdelatrisu/opsu/beatmap/BeatmapParser.java
sr229: I don't think you should use Collections in this case, as there would be instances they would derail out of order map-wise

src/itdelatrisu/opsu/beatmap/TimingPoint.java
itdelatrisu: Okay, that's a good reason.  Has this been tested with a beatLength of 0?  I'd be surprised if it doesn't break things (since I always assumed this would be positive in all of the code that uses this...).

src/itdelatrisu/opsu/beatmap/BeatmapParser.java
Lyonlancer5: The sorting is for "Crashpire" to load (and compute its star rating correctly), as according to its beatmap code:

```
[HitObjects]
425,152,250138,70,0,B|35556:20000|35556:20000|-16384:442,1,97280,4|0,0:0|0:0,0:0:0:0:
301,92,251370,54,0,L|262144:262144,1,194560
261,231,19692,6,0,P|306:209|329:150,2,116.5,0|0|2,0:0|0:0|0:0,0:0:0:0:
```

The first two objects are found at the end of the beatmap, and the "true" first note is actually the third (based on the timestamp in parameter two).
This _is_ unique to Aspire maps (as usually notes are placed in order of their time)

As for the Collections, would synchronizing prevent said instances?
I did only use them to create the hit object array.

Lyonlancer5:
About the CPU usage, in my case it only reaches 100% when the map in #297 loads the "ghost"
 sliders (experimental sliders enabled, HD skinning enabled):
screenshot_20171016_134757

> Comparing with the second/third picture in #297 is the change between `beatLength >= 0` and `beatLength > 0`

---------

and the map in #302 does the ~1500 combo sliders (both normal and experimental sliders):
screenshot_20171016_135030
screenshot_20171016_135033

src/itdelatrisu/opsu/beatmap/TimingPoint.java
Lyonlancer5: > I'd be surprised if it doesn't break things

It won't break normal maps. Ranked maps would never go down to these numbers.
So far, it hasn't broken the previous Aspire maps (Evanescent and Transform) in my tests.

-----
> Has this been tested with a `beatLength` of 0?

This is a snippet of #297's "source", [Helblinde - The Solace of Oblivion (Mapped by ProfessionalBox)](https://osu.ppy.sh/s/594751)

```
[TimingPoints]
...
57846,1E-298,4,2,1,60,1,0
...
113227,1E-298,4,1,1,75,1,0
113227,-100,4,1,1,75,0,1
116958,307.692307692308,4,1,1,75,1,0
118148,1E-298,4,1,1,75,1,0
...
```

Due to [`float`s minimum positive value](https://docs.oracle.com/javase/7/docs/api/java/lang/Float.html#MIN_NORMAL), the number `1E-298`, when parsed, would be considered a solid zero instead of `0.000[...]0001`. This would affect any operation that deals with multiplication, division or anything that would modify the number's precision.

The change in the conditional was intended to make it work like how osu-stable handles it.

Lyonlancer5:
Going to rework this and split fixes into two PRs, one for Crashpire and the other for the slider-related problems, to cleanup and use less commits for easier comparison, checking and testing.

PR #331, opened 2017-10-26
Display beatmap's source on song select

Lyonlancer5:
Updates the display on the song and main menus to display where the audio came from (e.g. `東方Project`, `Nico Nico Douga`, `ラブライブ! School idol project` etc.)

Expect use of Unicode characters as `beatmap.source` is not limited to English.

### DIFF
--- a/src/itdelatrisu/opsu/beatmap/Beatmap.java
+++ b/src/itdelatrisu/opsu/beatmap/Beatmap.java
@@ -399,11 +399,14 @@ public class Beatmap implements Comparable<Beatmap> {
 	}
 
 	/**
-	 * Returns a formatted string: "Artist - Title [Version]"
+	 * Returns a formatted string: "Source (Artist) - Title [Version]"
 	 * @see java.lang.Object#toString()
 	 */
 	@Override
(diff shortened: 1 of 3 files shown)


tpenguinltg:
Just curious, is this how osu! displays the source?

Perhaps this should be available behind a pref? I already know where my songs come from, so I, for one, don't really care to see the source, especially if it's full of kanji that I can't read.

Lyonlancer5:
song-select

Its close to how osu! displays the source in its user interface.

I had put the source display to maintain the completeness of the osu! UI even on different platforms, plus the fact that osu!'s search functionality also takes into account the beatmap source (to those who can read/type it)

If you want it behind a setting, I can put it just under the "Prefer metadata in original language" setting to maintain a sort-of right flow.

tpenguinltg:
Parity is good, which means if this were a setting, it should probably be on by default. I'd defer to @itdelatrisu for the placement of the setting (if he even wants one), but I agree that it would go there.

The code otherwise looks good.

itdelatrisu:
Merged, sorry for the terribly late response on this.

Issue #350, opened 2017-12-21
fd


Negative31:
[Type your description here. Feel free to delete the info below if it's not relevant.]

---
**Version:** 0.16.0
**Build date:** 2017-07-02 17:27
**OS:** Windows 7 (x86)
**JRE:** 1.8.0_121
**OpenGL version:** 4.5.13397 Compatibility Profile Context 15.200.1046.0 (ATI Technologies Inc.)

Lemmmy:
thank you

PR #352, opened 2017-12-26
[Technical] Detailed crash reports

Lyonlancer5:
### Overview

This pull request aims to provide more technical information to a crash report. 
May reduce the number of ambiguous errors (esp. the `Failed to load sound file` ones) and better distinguish details about it.

This system can be placed on deep sections of code which may be susceptible to errors.

### Sample
Generated by using a corrupted `applause.wav` on Windows.

```
---- opsu! Error Report ----

Time: 12/27/17 1:21 PM
Description: Failed to load audio file 'applause.wav'

javax.sound.sampled.UnsupportedAudioFileException: could not get audio input stream from input URL
	at javax.sound.sampled.AudioSystem.getAudioInputStream(AudioSystem.java:1155)
	at itdelatrisu.opsu.audio.SoundController.loadClip(SoundController.java:104)
	at itdelatrisu.opsu.audio.SoundController.init(SoundController.java:257)
	at itdelatrisu.opsu.states.Splash$2.run(Splash.java:172)


Additional details:
---------------------------------------------------------------------------------------

-- Resource Details --

	Skin Name: Default
	Audio File Name: applause.wav
	Audio File Extension: .wav
	Audio File Header (First 16 bytes): 0 0 0 0 56 78 5 0 57 41 56 45 66 6d 74 20 

-- Audio System Details --

	Audio Devices: 
	- Primary Sound Driver version Unknown Version (Unknown Vendor) ~ Direct Audio Device: DirectSound Playback
	- Speakers (Realtek High Definition Audio) version Unknown Version (Unknown Vendor) ~ Direct Audio Device: DirectSound Playback
	- Primary Sound Capture Driver version Unknown Version (Unknown Vendor) ~ Direct Audio Device: DirectSound Capture
	- Microphone (Realtek High Defini version Unknown Version (Unknown Vendor) ~ Direct Audio Device: DirectSound Capture
	- Port Speakers (Realtek High Definiti version 10.0 (Unknown Vendor) ~ Port Mixer
	- Port Microphone (Realtek High Defini version 10.0 (Unknown Vendor) ~ Port Mixer

-- System Information --

	Game Version: 0.16.0 - built on 2017-12-27 12:16
	Operating System: Windows 10 (amd64) version 10.0
	Java Version: 1.8.0_131, Oracle Corporation
	Java VM Details: Java HotSpot(TM) 64-Bit Server VM (mixed mode), Oracle Corporation
	OpenGL Version: 4.5.13447 Compatibility Profile Context 16.300.2201.0
	OpenGL Renderer: AMD Radeon (TM) R5 M430
	OpenGL Vendor: ATI Technologies Inc.
```

### Changes
- New report format (as seen from above)
- `SoundController`, `MusicController` and `GameImage` now generate a more defined report of what went wrong (e.g. the sound file name, current skin, audio devices.
- Full error reports are now uploaded to hastebin rather than posted alongside the GitHub issue when `Send Report` is clicked.


yugecin:
Have you tried pressing the report button?
I suspect it won't work too well as the whole report is too long to fit into the URL that is used to make a new issue on github, I ran into the same problem when I tried to add more info too.

One possible solution would be to first upload the dump to e.g. pastebin and then add the link to that dump in the github issue.

Lyonlancer5:
I had tested it without typing anything on the report, but so far it doesn't cut off.
However, due to your suspicion, I would revisit [`hastebin`](https://hastebin.com.about.md)'s [API](https://github.com/seejohnrun/haste-server/wiki/POST-api) to do just that.

I would have either used `pastie` (which is apparently down because haven't used it for a long while) and [`pastebin`](https://pastebin.com) but it seems that, for the time being, I'd have to stick with hastebin for now.

Lyonlancer5:
This concern:

> The amount of system information this dumps would probably be considered intrusive. Think more about user privacy with this -- for instance, prompt the user before sending the crash reports to a pastebin site (or just open the file and ask them to do it manually).

The error report is only uploaded to pastebin when the user clicks the `Send Report` button, otherwise, it is written out to a file separate from the log. It is entirely up to them whether they would want to send the report automatically or post it separately on a GitHub issue. 

In short, the upload is not automatic, it is with user consent.

Additionally, I had made it more comprehensive to determine the causes for repeated issues and some older (more obscure) ones, but if you (and others) do consider it intrusive, just tell me what to remove from the reporting system.

itdelatrisu:
Right, but doesn't the user only see the crash log contents if they *don't* auto-upload (unless I've misread the code)? I suppose most programs would do something similar for crash reports, but we'd also be sending the reports through a pastebin site, which I'm a bit uncomfortable with.

Here's an approach I'd take:
* "Send Report" sends you to the GitHub issues page **and** opens the crash log file.
* Change the crash popup text to also ask users to copy+paste the full crash log into the GitHub issue.
* Edit the issue template to include a section saying "Paste Crash Report Here".
We can wait for other opinions on this, too.

src/itdelatrisu/opsu/Container.java
```diff
@@ -93,14 +99,26 @@ public void start() throws SlickException {
...
+			ErrorHandler.error(report, true);
+			
 			e = null;
 			forceExit = true;
 		}
 
 		if (forceExit) {
 			Opsu.close();
-			System.exit(0);
+			System.exit(1);
```
itdelatrisu: Forgot to mention in the last review: this isn't necessarily an error (e.g. restarting for an update). You could do:
```
System.exit((e == null) ? 0 : 1);
```

src/itdelatrisu/opsu/Container.java
```diff
@@ -93,14 +99,26 @@ public void start() throws SlickException {
 
 		// report any critical errors
 		if (e != null) {
-			ErrorHandler.error(null, e, true);
+			ErrorReport report = new ErrorReport(null, e);
+			ErrorReportCategory info = new ErrorReportCategory("Game Engine Details");
+			info.addSection("State Info", new Callable<String>() {
```
itdelatrisu: Same comment as before -- isn't this just redundant info?

yugecin:
I just remember I saw uncomplete dumps when I tried stuffs, but the example dump you show in the first comment seems to work good when I last tried it.
Either way, [this stackoverflow post](https://stackoverflow.com/a/417184) suggests that URLs should preferably not exceed 2000 chars (also keep in mind you might need to urlencode).

I'd also like to point out that hastebin deletes pastes that are not viewed at least once in 30 days.

I think the process of users having to copy it themselves would be good. I have a little addition to add to your idea @itdelatrisu:
* When you press report, another window pops up
* That window only has a textfield with the dump and a button to copy the dump and open a new issue
* When you press it, the dump content should be copied to the clipboard if possible and create new issue as usual with the opsu version and default template that has been used up until now, with a big placeholder where the user should paste the dump that has been copied

It does require some manual process and [sometimes that is seemingly difficult](https://github.com/itdelatrisu/opsu/issues/350), but I guess it should be fine. Opening the crash dump log file might be too confusing or difficult as opposed to showing it in a new jframe like the error reporter.

tpenguinltg:
If we're looking for pastebin options, I would be in favour of using private [Gists](https://gist.github.com/)  because if the user already has a GitHub account to report issues, they should automatically have a Gist account.

src/itdelatrisu/opsu/Utils.java
```diff
@@ -132,9 +134,9 @@ public static void init(GameContainer container, StateBasedGame game) {
 		container.getInput().enableKeyRepeat();
 		container.setAlwaysRender(true);
 		container.setUpdateOnlyWhenVisible(false);
-
-		// record OpenGL version
-		ErrorHandler.setGlString();
+
```
tpenguinltg: Introduced whitespace?

src/itdelatrisu/opsu/crash/ErrorReport.java
```diff
@@ -0,0 +1,369 @@
...
+			os.flush();
+			os.close();
+
+			if (conn.getResponseCode() != 200)
+				return "Remote server responded with \'" + conn.getResponseMessage() + "\'.";
+
+			BufferedReader reader = new BufferedReader(new InputStreamReader(conn.getInputStream()));
+			String reply = reader.readLine();
+
+			// Parse the JSON response manually
```
tpenguinltg: The `org.json` package is available in this project. It would be preferable to use it instead of manually parsing it.

Lyonlancer5:
> I'd also like to point out that hastebin deletes pastes that are not viewed at least once in 30 days.

@yugecin It's implied that, once the issue is resolved, the crash log would be automatically deleted. 
Anyways, I do come and revisit a lot of the older issues here every now and then.

> In favor of using secret Gists

@tpenguinltg Something like the use of the [Gist API](https://developer.github.com/v3/gists) to post (and optionally, delete) the crash log? 
Or, maybe I'll use my API key from pastebin and set it to `Unlisted` **_then_** post it.

> Right, but doesn't the user only see the crash log contents if they _don't_ auto-upload.

@itdelatrisu  The crash log contents is the one being displayed in the error reporter.

tpenguinltg:
@Lyonlancer5 I forgot about authentication. Perhaps Gist isn't the best idea since they would all be anonymous without authentication, and I don't think it'd be reasonable to ask users to create API keys. It may be reasonable to ask for a username and password so we can get a [programmatically-acquired OAuth2 token](https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization), but then you run into the issue of two-factor authentication. At that point, though, we may as well deal with the [issues API](https://developer.github.com/v3/issues/#create-an-issue) directly.

Certainly do _not_ commit an API key into the repo, so anything that requires authentication via API key is out. Something like [ptpb.pw](https://ptpb.pw/) is one that doesn't require authentication, but I don't know how long they keep their pastes for.

I would be in against deleting crash logs as they may provide insights in future issues.

It seems, then, if there's no reasonable pastebin solution that preferably doesn't require API authentication and near-indefinite expiry time, then @itdelatrisu and @yugecin's manual method seems like the better option. To suggest an alternative to the text area for the user to copy, you could direct them to upload a file containing the crash report details; a summary goes in the issue body and extra details in the file.

Lyonlancer5:
> [ptpb.pw](https://ptpb.pw)

Based on the overview of their API, it's implied that pastes don't expire (unless the [sunset](https://ptpb.pw/#id15) flag is set). 

However, since most have decided for the manual method, then OK, that's what would be implemented.

Additional info: the user could simply drag the crash file itself to the issue post because files can be attached to the issue post/comments.

Lyonlancer5:
As it stands, the details provided by the new system is a bit intrusive and too technical.
Will revisit and re-implement in a different way.

Issue #362, opened 2018-02-12
NullPointerException when playing some videos


tpenguinltg:
A `NullPointerException` occurs when the `tbr` of a video is not completely numeric. Example video: [103482 - Rhodanthe* - Jumping!! (TV Size)](https://osu.ppy.sh/s/103482)

Here is the relevant line from FFmpeg output. Note that `90k tbr` does not match the regex in [FFmpeg.java:111](https://github.com/itdelatrisu/opsu/blob/6a3c6f78d3460f27685da957239464ba7d3eeeee/src/itdelatrisu/opsu/video/FFmpeg.java#L111), which looks for a number with an optional decimal point followed by "tbr", causing the regex utility to return `null`.
```
    Stream #0:0: Video: mjpeg, yuvj420p(pc, bt470bg/unknown/unknown), 854x480, 90k tbr, 90k tbn, 90k tbc
```

FFmpeg's output to stderr [is not meant to be parsed](https://trac.ffmpeg.org/ticket/5880#comment:3). The proper way to solve this is to use [FFprobe](https://trac.ffmpeg.org/wiki/FFprobeTips), with the flags given in this [Ask Ubuntu answer](https://askubuntu.com/a/468003). The output is printed to stdout and the result of the division gives you the frame rate.

```
$ ffprobe -v 0 -of csv=p=0 -select_streams v:0 -show_entries stream=r_frame_rate OP480.avi
90000/1
```

Note that we don't currently bundle `ffprobe`.

* * *
## Stack trace

```
Sun Feb 11 19:00:55 EST 2018 ERROR:null
java.lang.NullPointerException
        at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1838)
        at sun.misc.FloatingDecimal.parseFloat(FloatingDecimal.java:122)
        at java.lang.Float.parseFloat(Float.java:451)
        at itdelatrisu.opsu.video.FFmpeg.extractMetadata(FFmpeg.java:111)
        at itdelatrisu.opsu.video.Video.<init>(Video.java:80)
        at itdelatrisu.opsu.states.Game.loadVideo(Game.java:1976)
        at itdelatrisu.opsu.states.Game.enter(Game.java:1654)
        at org.newdawn.slick.state.StateBasedGame.update(StateBasedGame.java:248)
        at org.newdawn.slick.GameContainer.updateAndRender(GameContainer.java:702)
[...]

PR #363, opened 2018-02-12
Avoid code duplication in FFmpeg framerate parsing

tpenguinltg:
In case you don't get to it first. (Related: 0b36328, #362)

`(k?)` is used instead of `(k)?` so that the region will always match, even if it's empty.


tpenguinltg:
Closed by 6449b73

itdelatrisu:
Oops, didn't see this in time: 6449b73
Anyhow, you can't use `==` on strings. :P

tpenguinltg:
Oh, you're right. I've been spoiled by JavaScript. I guess it just evaluated to false and didn't do the multiplication.

PR #371, opened 2018-03-10
Add catch clause for UnsupportedAudioFileException and fix LineUnavailableExceptions 

Lyonlancer5:
## Overview
Needs more testing on other setups.
Adds more details with issues regarding sound file loading and possibly fix the Linux side of `LineUnavailableException`s.
Has been tested on Windows (`Windows 10 x64`) and Linux (`Mint Cinnamon 18.3 x64`) with no ill effects.

### Changes: 
- Don't open a line with the same format twice (as found in [this StackOverflow post](https://stackoverflow.com/questions/11915469/java-sound-format-not-supported))

- Add an additional catch clause for `UnsupportedAudioFileException` to better distinguish audio file-specific errors (such as corrupted files) and more general errors (such as non-present decoders for an audio type)


itdelatrisu:
Thanks! Will merge since it seems harmless enough.

PR #375, opened 2018-03-20
Add About entry to copy debug info

### LINKED ISSUE #374: Add a button to show/copy debug information
We should expose a button, probably in the same menu that the GitHub button shows, that causes a dialog to appear that shows the information returned by [`ErrorHandler#getEnvironmentInfoForIssue()`](https://github.com/itdelatrisu/opsu/blob/db376af25e0977ea9bd28a931a06e1387e6ee0e9/src/itdelatrisu/opsu/ErrorHandler.java#L192) for the user to copy. This is similar to having the user click "Report an Issue" and copying the information out of the textarea, but with the advantage that it doesn't require the user to be online and load a Web page just to get that information.

I imagine this would require some refactoring to extract the dialog out of `ErrorHandler`.

An alternative UI to consider: instead of opening a dialog, which might look out of place, just copy the information to the clipboard and display a notification that the info has been copied to the clipboard. The downside to this is that if the user wanted to simply view it without copying it, they'd have to paste the info in some text editor first. This approach would probably be simpler to implement, but I'm not sure which one is better UX-wise.

**Motivation:** implementing this will allow us to give straight-forward, cross-platform instructions to get the required information from their systems for issues like #373 and #370 where the submitter doesn't initially submit that info.

tpenguinltg:
Adding this gives us a reliable, cross-platform way of getting this info for issues that don't already have it.

I went with the clipboard implementation because it was the simplest to implement and it would be the most likely use case.

Resolves #374

### DIFF
--- a/src/itdelatrisu/opsu/states/ButtonMenu.java
+++ b/src/itdelatrisu/opsu/states/ButtonMenu.java
@@ -24,6 +24,7 @@ import itdelatrisu.opsu.GameMod;
 import itdelatrisu.opsu.Opsu;
 import itdelatrisu.opsu.OpsuConstants;
 import itdelatrisu.opsu.ScoreData;
+import itdelatrisu.opsu.Utils;
 import itdelatrisu.opsu.audio.MusicController;
 import itdelatrisu.opsu.audio.SoundController;
 import itdelatrisu.opsu.audio.SoundEffect;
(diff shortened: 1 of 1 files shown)


itdelatrisu:
Looks fine. Thanks!

Issue #401, opened 2018-06-11
Wrong stack leniency calculations, game falls into infinite loop at the beginning of TheFatRat - Mayday


Wieku:
Opsu! 0.16.1 has some problems with this 2B map: https://osu.ppy.sh/beatmapsets/756794#osu/1605148

Just after few seconds game falls into (probably) infinite loop in main thread, because game just freezes without any thrown exception: https://img.wieku.me/18/2018-06-11_17-47-48.png

yugecin:
This only happens with auto mod enabled, right?

Hitobject 5's endtime is the same as the starttime of hitobject 6 causing the `t` (last) parameter of following line to be `Infinity` because of division by zero, making the autopoint located at `(-Infinity, Infinity)`. At the end the game is kind of stuck at drawing the cursortrail because it's trying to draw cursortrails all the way to position `(-2147483648, 2147483647)`
https://github.com/itdelatrisu/opsu/blob/081e731d0cc7a9b40765b188867f555bbc4eb688/src/itdelatrisu/opsu/states/Game.java#L470

Wieku:
Yes, this happens only in automod.

And i just found that stack leniency calculations are wrong for this map:
opsu: https://img.wieku.me/18/screenshot_20180611_214301.png
osu: https://img.wieku.me/18/screenshot001osu.png
danser: https://img.wieku.me/18/1528746541477875300.png

PR #403, opened 2018-06-11
fix near-infinite loop on certain 2B maps

yugecin:
fixes issue #401 (minus the stack leniency calculations) caused by the cursor trail taking a very long time to draw, caused by autopoint being at infinity coordinates, caused by a division by zero caused by one hitobject's start time being the same as the previous hitobject's end time.

I've added an explicit check for `0`. Clamping the  `t` value between `[0, 1]` would've also fixed it, but I don't think that's necessary.


itdelatrisu:
Awesome, thanks for continuing to support the project! :)

PR #406, opened 2018-06-18
Fixed stack leniency calculations

Wieku:
Fixes the first part of #401 issue. Stack leniency calculations were wrong for sliders, which were clearly visible on some 2B maps, such as "Mayday" or "Endless Tewi-ma Park."

### DIFF
--- a/src/itdelatrisu/opsu/states/Game.java
+++ b/src/itdelatrisu/opsu/states/Game.java
@@ -2329,7 +2329,6 @@ public class Game extends BasicGameState {
 	private void calculateStacks() {
 		// reverse pass for stack calculation
 		for (int i = gameObjects.length - 1; i > 0; i--) {
-
 			HitObject hitObjectI = beatmap.objects[i];
 
 			// already calculated
@@ -2343,7 +2342,7 @@ public class Game extends BasicGameState {
 						continue;
 
 					// check if in range stack calculation
-					float timeI = hitObjectI.getTime() - (approachTime * beatmap.stackLeniency);
+					float timeI = hitObjectI.getTime() - approachTime * beatmap.stackLeniency;
 					float timeN = hitObjectN.isSlider() ? gameObjects[n].getEndTime() : hitObjectN.getTime();
 					if (timeI > timeN)
 						break;
@@ -2387,7 +2386,7 @@ public class Game extends BasicGameState {
 						continue;
 
 					// check if in range stack calculation
-					float timeI = hitObjectI.getTime() - (approachTime * beatmap.stackLeniency);
+					float timeI = hitObjectI.getTime() - approachTime * beatmap.stackLeniency;
 					float timeN = hitObjectN.getTime();
 					if (timeI > timeN)
 						break;
@@ -2402,7 +2401,6 @@ public class Game extends BasicGameState {
 					}
 				}
 			}
-
 		}
 
 		// update hit object positions


src/itdelatrisu/opsu/states/Game.java
```diff
@@ -2329,57 +2329,80 @@ else if (endTime - trackPosition < approachTime)
 	private void calculateStacks() {
 		// reverse pass for stack calculation
 		for (int i = gameObjects.length - 1; i > 0; i--) {
+
```
tpenguinltg: Not needed.

src/itdelatrisu/opsu/states/Game.java
```diff
@@ -2329,57 +2329,80 @@ else if (endTime - trackPosition < approachTime)
...
+							hitObjectI.getX(), hitObjectI.getY(),
+							hitObjectN.getX(), hitObjectN.getY()
+					);
+					if (distance < STACK_LENIENCE * HitObject.getXMultiplier()) {
+						hitObjectN.setStack(hitObjectI.getStack() + 1);
+						hitObjectI = hitObjectN;
+					}
 				}
 			}
+
```
tpenguinltg: Not needed.

src/itdelatrisu/opsu/states/Game.java
```diff
@@ -2329,57 +2329,80 @@ else if (endTime - trackPosition < approachTime)
...
 					}
 				}
+			} else if (hitObjectI.isSlider()) {
+				for (int n = i - 1; n >= 0; n--) {
+					HitObject hitObjectN = beatmap.objects[n];
+					if (hitObjectN.isSpinner())
+						continue;
+
+					// check if in range stack calculation
+					float timeI = hitObjectI.getTime() - (approachTime * beatmap.stackLeniency);
```
tpenguinltg: Parentheses around `approachTime * beatmap.stackLeniency` can be removed as `*` has precedence over `-`.

src/itdelatrisu/opsu/states/Game.java
```diff
@@ -2329,57 +2329,80 @@ else if (endTime - trackPosition < approachTime)
...
-							// hit object below slider end
-							if (distance < STACK_LENIENCE * HitObject.getXMultiplier())
-								hitObjectJ.setStack(hitObjectJ.getStack() - offset);
-						}
-						break;  // slider end always start of the stack: reset calculation
+						hitObjectN.setStack(hitObjectI.getStack() + 1);
+						hitObjectI = hitObjectN;
 					}
 				}
+			} else if (hitObjectI.isSlider()) {
```
tpenguinltg: Do we care about the case where the hit object is not a circle or slider (e.g. when it's a spinner)? If so, should we be doing anything with it in an `else` clause?

Wieku: No, we don't care about them.

src/itdelatrisu/opsu/states/Game.java
Wieku: I was basing on the code that was before that had parenthesis

tpenguinltg: Ah, then I would have kept it the way you had it, but it seems I didn't reply soon enough. Oh well, it's done now, so let's see if @itdelatrisu cares much about it.

Wieku:
I can't request a review (i don't know why), so summon @itdelatrisu

Issue #45, opened 2015-03-12
Visible Cursor


abalabahaha:
When a beatmap is loading after you click on it in the song menu, the mouse arrow becomes visible and stays visible until it is moved out of the window and back in.
Affects: OSX 10.10 + jdk1.8.0_31/jdk1.7.0_75

itdelatrisu:
Does this only happen on OSX?  (I don't own any Apple devices, but haven't been able to reproduce this on Windows.)

abalabahaha:
I cannot replicate this in Linux (Ubuntu either). Should be just OSX.

itdelatrisu:
Hm, okay.  I don't know where to start looking, then... and it's really a strange bug if it doesn't happen anywhere else.  I'll mark this as "help wanted", but I really doubt it's a problem with my code.  I didn't look in depth, but it seems like LWJGL has had a number of cursor-related bugs for OSX; these might have been fixed in newer versions...

Let me know if anything else is wrong on OSX, since I haven't run opsu! on a Mac since the initial release. x.x

abalabahaha:
Uncommenting states/Game.java line 842 fixes this, but that may have adverse side effects (can't move mouse out of window, whatever touchscreen issue the line above that talked about)...

itdelatrisu:
And this doesn't happen in any other state?  It doesn't make sense that `setMouseGrabbed()` would fix the issue since it's not being called anywhere else, so I wonder why it works...

I actually had the line there because osu! binds the mouse inside the window during gameplay -- that part is desirable behavior.  But for some reason it also completely disables touchscreen clicks for me, so I can't enable it as-is.

abalabahaha:
I remember it sometimes occurring in other states, but not very often.
I think it might have to do with tiny lag spikes (beatmap load) and the program temporarily relinquishing hold of the mouse... Just a guess.

itdelatrisu:
I guess you could call `setMouseGrabbed(true)` during every state's `enter()` method and then disable it immediately on the first `update()` call (or something like that), but it sounds like an awful hack.

Is it only happening in the game state due to the image loading on the first run?  (e.g. does the cursor appear anytime besides the first game?)  If so, it might not be too bad to grab the mouse before image loading and ungrab it afterwards -- at the beginning and end of the `loadImages()` method.  Could you try it?

abalabahaha:
[...]

PR #423, opened 2018-08-19
Add option to lock cursor to window

tpenguinltg:
Locking the cursor to the window prevents the mouse from leaving the window. This is useful in windowed mode where accidentally leaving the window results in the game being paused.

The `F9` key has been globally bound to toggle this option. osu! uses `F9` to toggle its extended chat window, but opsu! is unlikely to implement a chat feature. As well, it is close to the `F10` key, which toggles mouse buttons.


itdelatrisu:
Unfortunately this has been broken in Slick2D since the beginning: https://github.com/itdelatrisu/opsu/issues/45#issuecomment-78431654

Just tried it again, and this still completely breaks touchscreen support. :/

tpenguinltg:
I realized that a drawing tablet will probably experience the same issue as a touch screen, both being absolute pointing devices. I'll see what I can do with that.

kleines-L:
please work on opsu

PR #426, opened 2018-08-21
Fixed wrong insertion of Catmull curve's repeated first point

Wieku:


itdelatrisu:
Doesn't this not matter because the list is empty?

Issue #399, opened 2018-06-04
Edit slider (menu) values with arrow keys


takase1121:
Its easier to tune the offset this way.

tpenguinltg:
This would be nice! If I'm not mistaken, I think we'll need to introduce the concept of control focus first (i.e. tabbing through the controls). If we're doing this, we should make the rest of the controls keyboard-accessible as well ([space] to toggle checkboxes, etc.)

takase1121:
Probably, and thanks for noticing this request! I can't wait to see this in future version of opsu!

PR #429, opened 2018-08-29
Added ability to adjust options sliders with arrow keys

NickGuz:
You can use arrow keys to make +1 or -1 adjustments to options sliders (offset, volume, etc.)

If there's a better way of doing this I'll give it another go but this is functional at least for the time being.

### DIFF
--- a/src/itdelatrisu/opsu/options/OptionsOverlay.java
+++ b/src/itdelatrisu/opsu/options/OptionsOverlay.java
@@ -905,6 +905,15 @@ public class OptionsOverlay extends AbstractComponent {
 			}
 		}
 
+		// increment or decrement slider value with arrow keys
+		if (hoverOption != null && hoverOption.getType() == OptionType.NUMERIC) {
+			if (input.isKeyPressed(Input.KEY_RIGHT))
+				adjustSliderWithKey(Input.KEY_RIGHT);
+			else if (input.isKeyPressed(Input.KEY_LEFT))
+				adjustSliderWithKey(Input.KEY_LEFT);
+		} else
+			input.clearKeyPressedRecord();
+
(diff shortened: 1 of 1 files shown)


tpenguinltg:
Linking #399/

src/itdelatrisu/opsu/options/OptionsOverlay.java
```diff
@@ -1350,6 +1359,27 @@ private void adjustSlider(int mouseX, int mouseY) {
 		}
 	}
 
+	/**
+	 * Increments or decrements slider value if user presses right or left arrow key.
+	 * @param key the pressed key
+	 */
+	private void adjustSliderWithKey(int key) {
```
tpenguinltg: I think instead of passing in the key, you should pass in something more symbolic, like a delta value.

src/itdelatrisu/opsu/options/OptionsOverlay.java
```diff
@@ -1350,6 +1359,27 @@ private void adjustSlider(int mouseX, int mouseY) {
...
+		int oldSliderValue = hoverOption.getIntegerValue();
+
+		// increment or decrement value
+		if (key == Input.KEY_RIGHT) {
+			hoverOption.setValue(hoverOption.getIntegerValue() + 1);
```
tpenguinltg: You have `oldSliderValue` that holds `hoverOption.getIntegerValue()`; you may as well use it here instead of calling the getter again. Same goes for the line below.

tpenguinltg:
I just tested it. It appears to work as expected. It's a bit awkward having to hover over the slider option first, but it's the best we can do with what we have for now, and it's nice having such precise control over the value.

PR #462, opened 2018-12-14
Update Opsu.java

ev-ev:
:)


ev-ev:
Wait what, I never added any commits! I believe this is a mistake or something... Sorry. I may have misclicked or something :/

PR #491, opened 2019-07-01
Add "Perfect" Mod(WIP)

ghost:
Few issues with this:
A) Can't find the fitting mod icon, so i'm using the sudden death mod icon as a placeholder
B) Instead of instantly restarting the level like on the original osu, you will just fail instead(its better this way IMO)

### DIFF
--- a/src/itdelatrisu/opsu/GameData.java
+++ b/src/itdelatrisu/opsu/GameData.java
@@ -1656,11 +1656,14 @@ public class GameData {
 		boolean hideResult = (hitResult == HIT_300 || hitResult == HIT_300G || hitResult == HIT_300K) && !Options.isPerfectHitBurstEnabled();
 		
 		// If player didn't get a perfect hit, fail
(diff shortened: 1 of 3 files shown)


src/itdelatrisu/opsu/GameData.java
```diff
@@ -1655,7 +1654,13 @@ public void sendHitResult(
 			return;  // "relax" and "autopilot" mods: hide misses
 
 		boolean hideResult = (hitResult == HIT_300 || hitResult == HIT_300G || hitResult == HIT_300K) && !Options.isPerfectHitBurstEnabled();
-		hitResultList.add(new HitObjectResult(time, hitResult, x, y, color, hitResultType, curve, expand, hideResult));
+		
+		// If player didn't get a perfect hit, fail
+		if (GameMod.PERFECT.isActive() && (hitResult == HIT_300 || hitResult == HIT_300G || hitResult == HIT_300K)) {
```
ghost: Oops! I made a derp here, will fix in a short while

